PR #48, opened 2026-03-03
cli: add examples command and improve ui startup

shehan360:
## Summary

Add `arksim examples` CLI command to download example projects from the GitHub repo without cloning. Improve `arksim ui` startup by auto-loading scenarios from config.

## Changes

- Add `arksim examples` CLI command with three modes: download all, download one by name, or list available examples
- Auto-load scenarios in `arksim ui` when config has `scenario_file_path` (falls back to demo)
- Add directory scope hint in file browser modal
- Remove redundant project root path from sidebar header

Will add changelog entry as follow-up before merge.

## How to Test

- [x] `ruff check .` passes
- [x] `pytest tests/` passes
- [ ] Manual verification:
  - Run `arksim examples --list` and verify available examples are listed
  - Run `arksim examples bank-insurance` and verify it downloads to `./examples/bank-insurance/`
  - Run `arksim ui` from a directory with a config.yaml containing `scenario_file_path` and verify scenarios auto-load
  - Run `arksim ui` from an empty directory and verify the demo scenario loads
  - Open the file browser and verify the scope hint is shown

## Notes

Uses only stdlib modules (`urllib.request`, `tarfile`, `io`) — no new dependencies.

## Reviewers

### DIFF
--- a/tests/unit/test_error_detection.py
+++ b/tests/unit/test_error_detection.py
@@ -1,88 +1,9 @@
 # SPDX-License-Identifier: Apache-2.0
-"""Tests for error_detection: collect_agent_behavior_failure_reasoning.
+"""Tests for error_detection: collect_agent_behavior_failure_reasoning."""
 
-Uses importlib to load modules directly from file paths, bypassing
-the arksim.evaluator.__init__.py which pulls in heavy deps
-(langchain, azure, etc.) that may not be installed in the test env.
-"""
-
-import importlib.util
-import sys
-import types
-from pathlib import Path
-from unittest.mock import MagicMock
-
-# ---------------------------------------------------------------------------
-# Direct-import helpers: load .py files without triggering __init__.py chains
-# ---------------------------------------------------------------------------
-_ARKSIM_ROOT = Path(__file__).resolve().parents[2] / "arksim"
-
-
-def _load_module(name: str, filepath: Path) -> types.ModuleType:
-    """Load a Python module from filepath into sys.modules[name]."""
-    spec = importlib.util.spec_from_file_location(name, filepath)
-    mod = importlib.util.module_from_spec(spec)
-    sys.modules[name] = mod
-    spec.loader.exec_module(mod)
-    return mod
-
-
-# 1. Load lightweight dependency modules first
-_enums_mod = _load_module(
-    "arksim.evaluator.utils.enums",
-    _ARKSIM_ROOT / "evaluator" / "utils" / "enums.py",
-)
-
-_base_metric_mod = _load_module(
-    "arksim.evaluator.base_metric",
-    _ARKSIM_ROOT / "evaluator" / "base_metric.py",
-)
-
-# 2. Load entities.py directly
-_entities_mod = _load_module(
-    "arksim.evaluator.entities",
-    _ARKSIM_ROOT / "evaluator" / "entities.py",
-)
-
-# 3. Stub LLM deps before loading error_detection
-_orig_llms = sys.modules.get("arksim.llms")
-_orig_llms_chat = sys.modules.get("arksim.llms.chat")
-sys.modules["arksim.llms"] = MagicMock()
-sys.modules["arksim.llms.chat"] = MagicMock()
-
-_prompts_mod = _load_module(
-    "arksim.evaluator.utils.prompts",
-    _ARKSIM_ROOT / "evaluator" / "utils" / "prompts.py",
-)
-_schema_mod = _load_module(
-    "arksim.evaluator.utils.schema",
-    _ARKSIM_ROOT / "evaluator" / "utils" / "schema.py",
-)
-
-_error_detection_mod = _load_module(
-    "arksim.evaluator.error_detection",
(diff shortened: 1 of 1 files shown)


arksim/cli.py
```diff
@@ -189,6 +192,65 @@ def _run_show_prompts(category: str | None) -> None:
...
+        if list_only:
+            print("Available examples:")
+            for ex in available:
+                print(f"  {ex}")
+            return
+
+        if name and name not in available:
+            logger.error(f"Unknown example '{name}'. Available: {', '.join(available)}")
+            sys.exit(1)
+
```
itsarbit, line 231: The `filter=\"data\"` param is great for blocking path traversal, but it was only added in Python 3.12. Since we support 3.10+, this will either throw a TypeError or silently skip the filter on older versions. That's a CVE-2007-4559 risk.

Can we add a fallback? Something like:

```python
import sys
if sys.version_info >= (3, 12):
    tar.extract(member, ".", filter="data")
else:
    # manual path check
    if os.path.isabs(member.name) or ".." in member.name.split("/"):
        continue
    tar.extract(member, ".")
```

arksim/cli.py
```diff
@@ -189,6 +192,65 @@ def _run_show_prompts(category: str | None) -> None:
...
+        for member in tar.getmembers():
+            if member.name.startswith(filter_prefix):
+                member.name = member.name.removeprefix(f"{top}/")
+                tar.extract(member, ".", filter="data")
+
+    logger.info(f"Downloaded to {os.path.abspath(dest_path)}")
+
+
```
itsarbit, line 260: This routes through the generic key-value parser right? If `parse_extra_args` expects `--key value` pairs, a bare `--list` flag with no value probably won't parse correctly. Might be cleaner to add it as a dedicated argparse flag on the examples subcommand instead.

arksim/cli.py
```diff
@@ -189,6 +192,65 @@ def _run_show_prompts(category: str | None) -> None:
...
 def build_parser(valid_commands: list[str] | None = None) -> argparse.ArgumentParser:
     """Build the CLI argument parser.
```
itsarbit: Nit: `config_file` is the positional arg for YAML config paths. It works here by accident since it's just a positional string, but someone reading this later will be confused about why an example name is coming from `config_file`. Worth either renaming or adding a comment.

arksim/ui/frontend/app.js
```diff
@@ -176,7 +176,7 @@ function arksim() {
       this._connectWs();
       await this._fetchProjectRoot();
       await this._loadConfigs();
-      await this._loadDemoScenario();
```
itsarbit, line 179: UI changes look good. Auto-loading scenarios from config instead of always falling back to demo is a nice UX improvement. The scope hint on the file browser is helpful too.

arksim/cli.py
itsarbit: To add context on why CI didn't catch this: the test matrix does run 3.10-3.13, but there are no tests for `_run_examples` at all. So the `filter="data"` kwarg never actually gets exercised in CI.

Even if we had tests, the failure would only show up on the 3.10 and 3.11 runners (not 3.12/3.13), so it could easily get missed as a partial matrix failure.

Worth adding at least a basic test that mocks `urlopen` and hits the extract path. That would blow up immediately on the 3.10 runner and catch this.

shehan360: Done in `4881563`. The manual path check (`os.path.isabs` + `..` split check) now runs unconditionally on all versions, with `filter="data"` as defence-in-depth on 3.12+:

```python
if os.path.isabs(member.name) or ".." in member.name.split("/"):
    continue
if sys.version_info >= (3, 12):
    tar.extract(member, ".", filter="data")
else:
    tar.extract(member, ".")
```

shehan360: Added in `dd55d6e`. `TestRunExamples` has 4 tests covering list, unknown name, dest-exists, and path traversal. The path traversal test also asserts the correct `filter` kwarg per Python version:

```python
if sys.version_info >= (3, 12):
    assert call.kwargs.get("filter") == "data"
else:
    assert "filter" not in call.kwargs
```

So this will blow up on the 3.10/3.11 runners if the version gate regresses.

arksim/cli.py
shehan360: Good call — refactored to dedicated subparsers in `4881563`. `examples` now has a proper `--list` flag (`action="store_true"`) and a `name` positional, so it no longer goes through `parse_extra_args` at all.

arksim/cli.py
shehan360: Resolved by the subparser refactor in `4881563`. `examples`, `show-prompts`, and `ui` now have their own dedicated subparsers with properly named args (`name`, `--category`, `--port`). Only the config-based commands (`simulate`, `evaluate`, `simulate-evaluate`) use `config_file`.

arksim/ui/frontend/app.js
shehan360: Thanks! 🙏

PR #50, opened 2026-03-04
feat: add version display to CLI and web UI

kmchen-arklex:
- Add -v/--version flag to arksim CLI (prints version and exits)
- Add arksim --version to help examples and README CLI reference
- Handle --version in additional_args for config-based commands
- Add GET /api/version endpoint for web UI
- Display version next to "Arksim" in sidebar (e.g. v0.0.1)

## Summary

Adds version display to the CLI and web UI so users can see the installed Arksim version.

## Changes

- Add `-v` / `--version` flag to arksim CLI (prints version and exits)
- Add `arksim -v` and `arksim --version` to help examples
- Handle `--version` in `additional_args` for config-based commands (e.g. `arksim simulate config.yaml --version`)
- Add `GET /api/version` endpoint for the web UI
- Display version next to "Arksim" in the UI sidebar (e.g. v0.0.1)
- Update README CLI reference with `arksim --version`

## How to Test

- [ ] `ruff check .` passes
- [ ] `pytest tests/` passes
- [ ] Manual verification: Run `arksim -v`, `arksim --version`, and `arksim simulate config.yaml --version`; start `arksim ui` and confirm version appears in the sidebar next to "Arksim"

## Notes

Version is sourced from `arksim.__version__` (from `hatch-vcs` at build time).

## Reviewers

@shehan360 @xinyang-articulate @arklexai/arksim-maintainers @itsarbit @jackytungArticulateAI


arksim/cli.py
```diff
@@ -293,6 +294,8 @@ def build_parser() -> argparse.ArgumentParser:
         epilog=textwrap.dedent("""\
 
             Examples:
+              arksim -v
```
itsarbit: These do the same thing: listing both is a bit redundant. I'd just keep one line:
```
  arksim --version
```

arksim/cli.py
```diff
@@ -391,6 +402,11 @@ def main() -> None:
         launch_ui(port=args.port)
         return
 
+    # Handle --version when passed after command (e.g. arksim simulate config.yaml --version)
+    if "--version" in (args.additional_args or []):
+        print(f"arksim {__version__}")
+        return
```
itsarbit: This shouldn't be needed  `argparse` already handles `--version` at the top-level parser before subcommand dispatch. Try running `arksim simulate config.yaml --version` without this block, it should still work.

If there's an edge case where it doesn't, this has a different output format (`arksim 0.0.4`) vs the argparse one (`arksim 0.0.4` via `%(prog)s`). Would be better to keep one code path.

arksim/ui/app.py
```diff
@@ -44,6 +45,10 @@ async def lifespan(app: FastAPI) -> AsyncIterator[None]:
     app.include_router(fs_router, prefix="/api")
     app.include_router(ws_router, prefix="/api")
 
+    @app.get("/api/version")
```
itsarbit: This route is defined inside the `lifespan` context manager, but all other routes use `APIRouter` objects. Would be more consistent to put it on a router or at least register it on `app` outside of `lifespan`.

Also, since `__version__` is a constant, this could just be a plain dict on module level rather than an async endpoint, but either way works.

kmchen-arklex: now uses APIRouter but kept inside app.py. Will move to a separate module if it grows

PR #52, opened 2026-03-04
chore: code quality audit fixes

itsarbit:
## Summary

Comprehensive code quality audit covering security, bug fixes, dead code removal, and OSS hygiene improvements.

### Security
- Fix path traversal vulnerabilities in `routes_filesystem` and `routes_results` API endpoints
- Add input validation to prevent directory traversal attacks

### Bug fixes
- Guard against `None` endpoint in `AgentConfig.get_endpoint()`
- Raise `ValueError` instead of returning raw string on parse failure in Azure OpenAI provider
- Ensure `ThreadPoolExecutor` `max_workers >= 1` in evaluator (required on Python 3.12+)
- Validate `num_workers` is a positive integer in workers utility
- Fix CLI arg parsing to allow negative number values (e.g. `-0.5`)
- Remove unreachable dead code branch in `evaluate.py`

### Code quality
- Convert `BaseAgent` to proper ABC with `@abstractmethod`
- Remove unused `flip_hist_content_only`, `LLMConfig`, `UNIQUE_BUGS` enum
- Strip 14 unused error message constants
- Replace magic numbers with named constants (`SCORE_NOT_COMPUTED`, `BEHAVIOR_FAILURE_THRESHOLD`)
- Fix dynamic version import from hatch-vcs

### Documentation and OSS hygiene
- Fix GitHub Actions versions (`@v6` does not exist, corrected to `@v4`/`@v5`)
- Align coverage threshold (pyproject.toml and CI both at 60%)
- Add SPDX Apache-2.0 license headers to all example files
- Add missing `__init__.py` files for proper packaging
- Add pre-commit hooks (`check-merge-conflict`, `detect-private-key`, `check-json`)
- Fix broken README table row, typos in CONTRIBUTING.md and HTML report
- Add `.claude/` to `.gitignore`

### Test updates
- Update tests for ABC `BaseAgent` (concrete subclass), removed enum member, and stricter validation

## Test plan
- [x] All 320 tests pass (3 skipped for optional deps)
- [x] All pre-commit hooks pass (ruff, ruff-format, trailing whitespace, etc.)
- [x] Coverage remains above 60% threshold

PR #53, opened 2026-03-04
fix(examples): improve sample scenario files

xinyang-articulate:
## Summary

Add standalone simulation/evaluation config files for both examples, update the e-commerce example with A2A protocol support, and cap auto-resolved worker counts to avoid API rate-limit saturation.

## Changes

- Add separate config_simulate.yaml and config_evaluate.yaml for both the bank-insurance and e-commerce examples, enabling arksim simulate and arksim evaluate to run independently (not just simulate-evaluate)
- Add A2A protocol support to the e-commerce example: new agent_server/a2a/ module with AgentExecutor and Starlette-based server, plus config_a2a.yaml
- Introduce MAX_WORKERS = 8 cap in arksim/utils/concurrency/workers.py so "auto" no longer spawns hundreds of threads that stall on API rate-limits
- Update the evaluator to use the shared resolve_num_workers helper with the new cap
- Rename config_simulate_evaluate.yaml → config.yaml in the arksim package
- Remove config_simulate.yaml from .gitignore (now a tracked config file)
- Update e-commerce README.md, scenarios.json, docs, and requirements.txt to reflect A2A support and improved scenario descriptions

- [ ] Added an entry to `CHANGELOG.md` under the `[Unreleased]` section
- [x] Updated `README.md` (if installation, quickstart, or usage changed)

## How to Test

- [x] `ruff check .` passes
- [x] `pytest tests/` passes
- [x]  Manual verification: Run arksim simulate and arksim evaluate independently with the new config files for both bank-insurance and e-commerce examples. Verify the A2A agent server starts and responds correctly via config_a2a.yaml. Confirm that num_workers: auto resolves to at most 8 workers.

## Notes

- The MAX_WORKERS constant is set to 8 — high enough for I/O-bound LLM calls, low enough to avoid rate-limit saturation. This can be tuned later if needed.

## Reviewers

### DIFF
--- a/arksim/evaluator/evaluator.py
+++ b/arksim/evaluator/evaluator.py
@@ -16,7 +16,6 @@ from tqdm import tqdm
 from arksim.llms.chat import LLM
 from arksim.scenario import Scenarios
 from arksim.simulation_engine import Conversation, Simulation
-from arksim.utils.concurrency import resolve_num_workers
 from arksim.utils.output import load_json_file, save_json_file
 
 from .base_metric import ChatMessage, QualitativeMetric, QuantitativeMetric
@@ -130,9 +129,12 @@ class Evaluator:
 
         logger.info(f"Preprocessing complete: {total_turns} total turns to evaluate")
 
-        num_workers = resolve_num_workers(
-            self.params.num_workers, total_turns + len(conversations)
-        )
+        # When auto: enough workers for all turns + goal_completion calls at once.
+        # When explicit int: use that value as-is.
+        if self.params.num_workers == "auto":
+            num_workers = total_turns + len(conversations)
+        else:
+            num_workers = self.params.num_workers
 
         post_processing_steps = 1  # detect errors
         pbar = tqdm(
@@ -159,9 +161,8 @@ class Evaluator:
             with ThreadPoolExecutor(
                 max_workers=min(num_workers, len(all_turn_tasks))
             ) as executor:
-                # Inner parallelism: 0 means "one thread per metric" (bounded
-                # only by the metric count, which is small).  When the user
-                # gave an explicit int we reuse it as a cap for inner threads.
+                # auto → unlimited inner parallelism (0 = unlimited).
+                # explicit int → metrics capped at min(num_workers, num_metrics).
                 inner_workers = 0 if self.params.num_workers == "auto" else num_workers
                 turn_futures = {
                     executor.submit(


docs/e-commerce-customer-service-agent-evaluation.mdx
```diff
@@ -100,17 +100,17 @@ This agent is a RAG-based shopping assistant built with OpenAI Agents SDK, backe
 
         ```bash
         export A2A_API_KEY=<YOUR_A2A_API_KEY>
-        python -m examples.e_commerce.agent_server.a2a.server
+        python -m examples.e-commerce.agent_server.a2a.server
```
itsarbit, line 109: This will break. `examples.e-commerce` is not a valid Python module path - hyphens aren't allowed in Python identifiers. `python -m examples.e-commerce.agent_server.a2a.server` will fail.

Either revert to the underscore form (`examples.e_commerce`) or switch to running the file directly:
```bash
python examples/e-commerce/agent_server/a2a/server.py
```
Same issue on the Chat Completions command a few lines below.

examples/e-commerce/agent_server/a2a/agent_executor.py
```diff
@@ -0,0 +1,34 @@
...
+
+        user_input = context.get_user_input()
+        if not user_input or user_input.strip() == "":
+            raise ValueError("User input is required")
+
+        result = await agent.invoke(user_input)
+        await event_queue.enqueue_event(new_agent_text_message(result, context_id))
+
+    async def cancel(self, context: RequestContext, event_queue: EventQueue) -> None:
+        raise Exception("cancel not supported")
```
itsarbit: `raise Exception` is too broad here. Use `NotImplementedError` - that's the standard for unimplemented interface methods:
```python
async def cancel(self, context: RequestContext, event_queue: EventQueue) -> None:
    raise NotImplementedError("cancel not supported")
```

examples/e-commerce/agent_server/a2a/agent_executor.py
```diff
@@ -0,0 +1,34 @@
+from a2a.server.agent_execution import AgentExecutor, RequestContext
```
itsarbit, line 4: Missing `# SPDX-License-Identifier: Apache-2.0` header and `from __future__ import annotations`. Both are required by project convention (every other source file has them).

examples/e-commerce/agent_server/a2a/server.py
```diff
@@ -0,0 +1,109 @@
+import os
```
itsarbit, line 4: Same here - missing `# SPDX-License-Identifier: Apache-2.0` and `from __future__ import annotations`.

examples/e-commerce/agent_server/a2a/server.py
```diff
@@ -0,0 +1,109 @@
...
+    APIKeySecurityScheme,
+    In,
+    SecurityScheme,
+)
+
+from .agent_executor import (
+    EcommerceAgentExecutor,
+)
+
+API_KEY = os.environ.get("A2A_API_KEY", "")
```
itsarbit: If `A2A_API_KEY` isn't set, this defaults to `""`. A request with an empty `api-key: ""` header would then pass the `api_key != API_KEY` check (both are empty strings). Consider defaulting to `None` and checking explicitly:
```python
API_KEY = os.environ.get("A2A_API_KEY")
# ...
if not API_KEY or api_key != API_KEY:
    raise AuthenticationError("Invalid or missing API key")
```
Not critical for an example server, but sets a better pattern.

examples/e-commerce/agent_server/a2a/server.py
```diff
@@ -0,0 +1,109 @@
...
+        name="E-commerce Customer Service Agent",
+        description=(
+            "A RAG-powered e-commerce customer service agent that "
+            "provides concise, accurate answers based on product documentation."
+        ),
+        url=os.getenv("A2A_SERVER_URL", "http://localhost:9999/"),
+        version="1.0.0",
+        default_input_modes=["text"],
+        default_output_modes=["text"],
+        capabilities=AgentCapabilities(streaming=True),
```
itsarbit: The agent card advertises `streaming=True` but `EcommerceAgentExecutor.execute()` does a single `await agent.invoke()` and enqueues one complete message - it doesn't actually stream. Should this be `streaming=False`?

arksim/config_simulate.yaml
```diff
@@ -0,0 +1,51 @@
...
+simulated_user_prompt_template: null
+
+# LLM model
+model: gpt-5.1
+
+# LLM provider
+provider: openai
+
+# Workers for parallel processing; "auto" uses all available threads
+num_workers: auto
```
itsarbit: This comment is now outdated given the `MAX_WORKERS = 8` cap introduced in this same PR. "auto" no longer uses *all* available threads. Something like:
```yaml
# Workers for parallel processing; "auto" picks a suitable count (capped at 8)
```
Same applies to every other config file in this PR that has this comment.

arksim/utils/concurrency/workers.py
```diff
@@ -1,6 +1,11 @@
 # SPDX-License-Identifier: Apache-2.0
 from __future__ import annotations
 
+# Upper bound for 'auto' worker counts.  High enough to keep I/O-bound LLM
+# calls well-utilised, low enough to avoid saturating API rate-limits or
```
itsarbit: Heads up: this file was also modified in PR #52 (positive integer validation). There will be a textual merge conflict. The changes are semantically compatible so resolution should be straightforward, just be aware.

examples/e-commerce/agent_server/a2a/server.py
itsarbit: I'll change the code to enfore this

examples/e-commerce/agent_server/a2a/agent_executor.py
xinyang-articulate: @itsarbit just want to confirm, shall we add `# SPDX-License-Identifier: Apache-2.0` to everyone Python file and `from __future__ import annotations` to every non-empty one?

docs/e-commerce-customer-service-agent-evaluation.mdx
xinyang-articulate: I tested this and this should a valid command.

itsarbit: Just tested this on Python 3.13:

```
python3 -m examples.e-commerce.agent_server.a2a.server
Error while finding module specification for 'examples.e-commerce.agent_server.a2a.server'
(ModuleNotFoundError: No module named 'examples.e-commerce.agent_server.a2a')
```

Hyphens are not valid in Python module identifiers. `python -m` treats each segment as a Python identifier and `e-commerce` is parsed as `e` minus `commerce`. It will fail on any standard Python installation.

The fix is to use `python examples/e-commerce/agent_server/a2a/server.py` (file path) instead of `python -m` (module path).

examples/e-commerce/agent_server/a2a/agent_executor.py
itsarbit: Yes, both are required by project convention. PR #56 (currently open) adds automated enforcement for this:
- `insert-license` pre-commit hook for SPDX headers
- Ruff `FA100` rule for `from __future__ import annotations`

For this PR, just add both to any new Python files. Once #56 merges, pre-commit will catch missing headers/imports automatically.

docs/e-commerce-customer-service-agent-evaluation.mdx
xinyang-articulate: That's strange, I also tested on Python3.13 with conda but works fine. And there will be an error if `e_commerce` is used.

image.

xinyang-articulate: Will try to find a better solution for this problem

examples/e-commerce/README.md
```diff
@@ -21,12 +20,12 @@ Steps to run:
 
 3. From this example directory, run:
```
itsarbit, line 21: This is still `python -m examples.e-commerce.agent_server.a2a.server` which will fail (hyphens in Python module paths). Change to:
```bash
python examples/e-commerce/agent_server/a2a/server.py
```
Same fix needed for the Chat Completions command below.

arksim/scenario/entities.py
```diff
@@ -1,7 +1,7 @@
 # SPDX-License-Identifier: Apache-2.0
 from __future__ import annotations
 
-from pydantic import BaseModel, model_validator
+from pydantic import BaseModel, Field, model_validator
```
itsarbit, line 5: Making `knowledge` and `origin` optional is a schema change that affects all downstream consumers (simulation engine, evaluator, HTML report). Please add a test in `test_scenario_entities.py` that constructs a `Scenario` without these fields to confirm it validates, and add a CHANGELOG entry under Changed.

itsarbit (dismissed):
Looks good. Previous comments addressed. Approving.

PR #54, opened 2026-03-04
fix: update logic for resolving file paths

KathyLiu20:
## Summary

When resolving the scenario_file_path and simulation_file_path in the config.yaml, they are usually the relative path to the config file path. 
This PR modified the logic when validating the input models.
If paths are read from the config.yaml, we resolve the paths as config-relative paths.
If paths are overridden from cli, we resolve the paths as cwd-relative paths.

## Changes

In the model validation for SimulationInput and EvalutionInput, add the logic to check if the path variables are loaded from config.yaml or override from cli and resolve the path accordingly.

No docs needed, the user behavior won't change. This change make our input validation more robust.

## How to Test

- [x] `ruff check .` passes
- [x] `pytest tests/` passes
- [x] Manual verification:  run with config-relative scenario file path in config.yaml
- [x] run with cli override (cwd-relative path)

## Reviewers

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@
 ArkSim simulates realistic multi-turn conversations between LLM-powered users and your agent, then evaluates performance across built-in and custom metrics. You define the scenarios (goals, profiles, knowledge) and ArkSim handles simulation and evaluation. Works with any agent that exposes a Chat Completions API or A2A protocol endpoint.
 
 <p align="center">
-  <img src="docs/assets/arksim-flow.svg" alt="ArkSim flow: Scenarios → Simulation → Evaluation → Reports" width="100%">
+  <img src="https://raw.githubusercontent.com/arklexai/arksim/main/docs/assets/arksim-flow.svg" alt="ArkSim flow: Scenarios → Simulation → Evaluation → Reports" width="100%">
 </p>
 
 ### Why ArkSim?


itsarbit (commented):
Solid fix for a real pain point. The cwd-first, config-relative-fallback approach makes sense. A few things below.

arksim/simulation_engine/entities.py
```diff
@@ -62,6 +63,29 @@ def validate_simulation_input(self, info: ValidationInfo) -> Self:
...
+        if (
+            self.scenario_file_path
+            and config_path
+            and "scenario_file_path" not in cli_overrides
+            and not os.path.exists(self.scenario_file_path)
+        ):
+            config_dir = os.path.dirname(config_path)
+            scenario_path_relative_to_config = os.path.join(
+                config_dir, self.scenario_file_path
+            )
```
itsarbit, line 82: The fallback logic in SimulationInput and EvaluationInput is almost identical. Worth extracting a helper like `_resolve_config_relative_path(path, config_dir, cli_overrides, attr_name)` to avoid the duplication. If one gets a bug fix later the other will be forgotten.

arksim/evaluator/entities.py
```diff
@@ -80,6 +81,28 @@ def validate_evaluation_input(self, info: ValidationInfo) -> Self:
...
+            for attr in ("scenario_file_path", "simulation_file_path"):
+                path = getattr(self, attr)
+                if path and attr not in cli_overrides and not os.path.exists(path):
+                    config_relative = os.path.join(config_dir, path)
+                    if os.path.exists(config_relative):
+                        setattr(self, attr, config_relative)
+                        used_config_relative = True
+            if (
+                used_config_relative
+                and self.output_dir
```
itsarbit: This is subtle. If `used_config_relative` is true, output_dir also gets rewritten to config-relative even if the user didn't set it explicitly. That could be surprising if someone has a shared output directory they always use from cwd. Maybe worth a debug log line here so users can see what happened when paths get rewritten?

arksim/simulation_engine/entities.py
```diff
@@ -1,6 +1,7 @@
 # SPDX-License-Identifier: Apache-2.0
 from __future__ import annotations
 
+import os
 import sys
```
itsarbit, line 5: This fallback logic has a few edge cases (both paths exist, config_path is None, cli override for one but not the other). Would be good to add a few unit tests for the validator with mocked file existence. The context passing pattern is easy to test with `model_validate(..., context={...})`.

arksim/config/utils.py
```diff
@@ -39,3 +39,20 @@ def replace_env_var(match: re.Match[str]) -> str:
...
+    config_dir: str,
+    cli_overrides: set,
+    attr_name: str,
+) -> str | None:
+    """Return config-relative path for config-sourced paths, None for CLI-sourced.
+
+    Paths from config.yaml are resolved relative to the config file's directory.
+    Paths provided via CLI (in cli_overrides) are left as-is (cwd-relative).
+    Absolute paths pass through unchanged.
+    """
```
itsarbit, line 55: The docstring says "Absolute paths pass through unchanged" but there is no explicit check for that. It works by coincidence since `os.path.join("/config/dir", "/absolute/path")` returns the absolute path, but it would be clearer and safer to make that explicit:

```python
if os.path.isabs(path) or attr_name in cli_overrides:
    return None
```

arksim/evaluator/entities.py
```diff
@@ -76,10 +81,39 @@ class EvaluationInput(BaseModel):
...
+                        path, config_dir, cli_overrides, attr
+                    )
+                    if resolved is not None and resolved != path:
+                        logger.debug("%s resolved to: %s", attr, resolved)
+                        setattr(self, attr, resolved)
+
+            if self.custom_metrics_file_paths:
+                resolved_list = []
+                for p in self.custom_metrics_file_paths:
+                    resolved = resolve_config_relative_path(
```
itsarbit: This path resolution block is nearly identical to the one in `SimulationInput.validate_simulation_input`. Consider extracting a shared helper like:

```python
def resolve_model_paths(model, attrs: tuple[str, ...], info: ValidationInfo) -> None:
    config_path = info.context and info.context.get("config_path")
    cli_overrides = (info.context and info.context.get("cli_overrides")) or set()
    if not config_path:
        return
    config_dir = os.path.dirname(config_path)
    for attr in attrs:
        path = getattr(model, attr)
        if path:
            resolved = resolve_config_relative_path(path, config_dir, cli_overrides, attr)
            if resolved is not None and resolved != path:
                logger.debug("%s resolved to: %s", attr, resolved)
                setattr(model, attr, resolved)
```

Both validators could then call this with their respective attribute tuples.

arksim/evaluator/entities.py
```diff
@@ -76,10 +81,39 @@ class EvaluationInput(BaseModel):
...
+                        p, config_dir, cli_overrides, "custom_metrics_file_paths"
+                    )
+                    resolved_list.append(resolved if resolved is not None else p)
+                if resolved_list != self.custom_metrics_file_paths:
+                    logger.debug(
+                        "custom_metrics_file_paths resolved to: %s", resolved_list
```
itsarbit: `custom_metrics_file_paths` is checked as a single CLI override key, so if the user overrides it via CLI, all paths in the list are left as-is. This is probably correct, but worth a brief comment explaining that the override granularity is per-field, not per-element.

arksim/simulation_engine/entities.py
```diff
@@ -62,6 +67,26 @@ def validate_simulation_input(self, info: ValidationInfo) -> Self:
...
+            for attr in (
+                "scenario_file_path",
+                "output_file_path",
+                "agent_config_file_path",
+            ):
+                path = getattr(self, attr)
+                if path:
+                    resolved = resolve_config_relative_path(
+                        path, config_dir, cli_overrides, attr
+                    )
```
itsarbit: nit: Using `setattr` inside a `model_validator(mode="after")` works today, but if the model ever gets `frozen=True` in its `ConfigDict`, this will break. Not a blocker, just something to keep in mind.

tests/unit/test_simulation_input.py
```diff
@@ -27,3 +30,63 @@ def test_skip_validation_context(self) -> None:
...
+        si = SimulationInput.model_validate(
+            self._base_data(),
+            context=self._ctx(tmp_path),
+        )
+        assert si.scenario_file_path is None
+        assert si.output_file_path == str(tmp_path / "simulation.json")
+
+    def test_cli_override_prevents_output_resolution(self, tmp_path: Path) -> None:
+        """output_file_path stays as-is when set via CLI."""
+        si = SimulationInput.model_validate(
```
itsarbit, line 77: This test shows that the default `output_file_path` value (`"simulation.json"`) also gets config-relative resolution. Is that intentional? If someone relies on the default, the output would land next to the config file rather than in cwd. This is a behavior change for existing users who run from a different directory than where their config lives.

itsarbit (changes requested):
Overall the approach is solid and the tests cover the key cases well. A few things to address before merging.

arksim/simulation_engine/entities.py
shehan360: Done — extracted `resolve_model_paths()` into `arksim/config/utils.py`. Both validators now call it with their respective attribute tuples.

arksim/evaluator/entities.py
shehan360: Added `logger.debug` lines in `resolve_model_paths()` that log whenever a path gets rewritten.

arksim/simulation_engine/entities.py
shehan360: Added unit tests covering config-relative resolution, CLI overrides, absolute paths, missing config_path, default output path resolution, and list path attributes. See `test_simulation_input.py` and `test_evaluation_input.py`.

arksim/config/utils.py
shehan360: Done — added an explicit `os.path.isabs(path)` guard in `resolve_config_relative_path` that returns `None` early for absolute paths.

arksim/evaluator/entities.py
shehan360: Done — extracted `resolve_model_paths()` in `arksim/config/utils.py` that handles both single-value and list-of-paths attributes. Both validators now use it.

arksim/evaluator/entities.py
shehan360: Added a comment in `resolve_model_paths()`: "List path overrides are per-field, not per-element: if the CLI overrides the field, all elements are left as cwd-relative."

arksim/simulation_engine/entities.py
shehan360: Switched to `object.__setattr__()` which bypasses Pydantic's frozen check. This is the same pattern Pydantic uses internally in validators.

tests/unit/test_simulation_input.py
shehan360: Yes, intentional. All paths from config.yaml (including defaults) resolve relative to the config file's directory, so output lands next to the config when running `arksim --config /some/dir/config.yaml` from a different cwd. CLI overrides (`--output-file-path ./my_output.json`) remain cwd-relative for users who want that behavior.

README.md
```diff
@@ -21,9 +21,9 @@
 
 ---
 
-<p align="center">
-  <em>Demo video coming soon</em>
-</p>
+
```
itsarbit: The demo video iframe change is unrelated to the path resolution fix. Should be in a separate PR to keep the history clean.

arksim/simulation_engine/entities.py
```diff
@@ -62,9 +64,24 @@ def validate_simulation_input(self, info: ValidationInfo) -> Self:
...
+            resolve_model_paths(
+                self,
+                path_attrs=(
+                    "scenario_file_path",
+                    "output_file_path",
+                    "agent_config_file_path",
+                ),
+                list_path_attrs=(),
+                config_dir=os.path.dirname(config_path),
+                cli_overrides=(info.context and info.context.get("cli_overrides"))
```
itsarbit, line 80: The `skip_input_dir_validation` context check was removed. Previously the `simulate+evaluate` pipeline passed `skip_input_dir_validation: True` so that `EvaluationInput` wouldn't require `agent_config` at construction time. Now the pipeline path in `cli.py` no longer passes this context. Is that intentional? If the pipeline always has `simulation_file_path` set at that point, it should be fine, but worth double-checking that nothing else relied on `skip_input_dir_validation`.

arksim/cli.py
```diff
@@ -153,6 +153,23 @@ def _parse_value(value: str) -> bool | int | float | str:
...
+            continue
+        val = overrides[key]
+        if isinstance(val, str):
+            overrides[key] = [v.strip() for v in val.split(",")]
+        elif not isinstance(val, list):
+            overrides[key] = [val]
+
+
 def validate_overrides(overrides: dict, valid_keys: set) -> None:
     invalid_keys = set(overrides.keys()) - valid_keys
```
itsarbit, line 183: New `_coerce_list_overrides` helper looks useful but has a subtle issue: `get_origin(annotation)` won't return `list` when the field uses `Optional[list[str]]` or `list[str] | None`. In those cases `get_origin` returns `Union` (or `types.UnionType`), not `list`. You'd need to unwrap the Optional first.

Also, this function is doing type introspection at runtime on every CLI invocation. Consider adding a test for list-type fields with Optional annotations to make sure this works correctly.

arksim/cli.py
```diff
@@ -153,6 +153,23 @@ def _parse_value(value: str) -> bool | int | float | str:
...
+
+    for key in list(overrides):
+        if key not in model_cls.model_fields:
+            continue
+        annotation = model_cls.model_fields[key].annotation
+        if get_origin(annotation) is not list:
+            continue
+        val = overrides[key]
+        if isinstance(val, str):
+            overrides[key] = [v.strip() for v in val.split(",")]
```
itsarbit, line 177: The comma-splitting logic (`val.split(",")`) could be surprising for paths that contain commas (unlikely but possible). Worth a brief comment noting this convention.

itsarbit (changes requested):
Previous feedback is well addressed. The shared helper, explicit `isabs` check, and `object.__setattr__` are all good. A few new things to look at.

README.md
shehan360: Good catch. The iframe change was from a prior commit on main that got merged into this branch. I'll revert it from this PR.

arksim/simulation_engine/entities.py
shehan360: Intentional. `skip_input_dir_validation` was only checked by `SimulationInput`, not `EvaluationInput`. In the simulate+evaluate pipeline, `EvaluationInput` never needed it — it has no `agent_config` requirement. For `SimulationInput`, every real code path always provides `agent_config_file_path`, so the skip was dead code. Confirmed by running the full test suite.

arksim/cli.py
shehan360: Good point about `Optional[list[str]]`. Checked both models — all current list fields use plain `list[str]` (not optional), so `get_origin` correctly returns `list`. But I'll add a guard to unwrap `Optional`/`Union` for future safety, plus a test for it.

arksim/cli.py
shehan360: Fair point. Will add a comment noting the comma-splitting convention.

README.md
shehan360: Reverted — restored "Demo video coming soon" placeholder. The iframe change will go in a separate PR.

arksim/simulation_engine/entities.py
shehan360: Intentional. `skip_input_dir_validation` was only checked in `SimulationInput`, not `EvaluationInput` — so the `cli.py` context that passed it to `EvaluationInput` was already dead code. For `SimulationInput`, every real code path always provides `agent_config_file_path`, so the skip was unreachable. Full test suite passes without it.

arksim/cli.py
shehan360: Fixed — `_coerce_list_overrides` now unwraps `Optional`/`Union` types before checking for `list`. Added `test_optional_list_field_coerced` that verifies a `list[str] | None` field is correctly detected and coerced.

arksim/cli.py
shehan360: Added a comment: `# Comma-separated values are split into a list (e.g. "a.py,b.py" → ["a.py", "b.py"]).'

PR #55, opened 2026-03-04
fix: respect num_conversations_per_scenario in simulator

nrajanee:
## Summary
The simulator ignored num_conversations_per_scenario, producing only 1 conversation per scenario instead of the configured count.

## Changes
Added an inner for _ in range(self.simulator_params.num_convos_per_scenario) loop in Simulator.simulate() so each scenario spawns the correct number of conversations (arksim/simulation_engine/simulator.py).

Added a unit test for this. 

- [ x] Added an entry to `CHANGELOG.md` under the `[Unreleased]` section
- [x ] Updated relevant docs in `docs/` (if behavior, config, or API changed)
- [x ] Updated `README.md` (if installation, quickstart, or usage changed)
- [x ] No docs or changelog needed (explain why below)

## How to Test

Follow these steps and verify it's 6 conversations now. 

Screenshot 2026-03-04 at 12 31 06 PM

## Reviewers
@xinyang-articulate @shehan360 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,7 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-### Fixed 
+### Fixed
 
 - Simulator now respects `num_conversations_per_scenario`, previously only 1 conversation was generated per scenario
 


itsarbit (commented):
Good catch and nice that you included a test. Few things below.

arksim/simulation_engine/simulator.py
```diff
@@ -273,30 +273,31 @@ def on_turn_complete() -> None:
...
+                    done, running_tasks = await asyncio.wait(
+                        running_tasks, return_when=asyncio.FIRST_COMPLETED
+                    )
+                    for task in done:
+                        try:
+                            result = task.result()
+                            if result is not None:
+                                results.append(result)
+                        except Exception as e:
+                            logger.error(f"Error processing conversation: {str(e)}")
```
itsarbit, line 287: Each conversation spawned for the same scenario will share the same scenario_id, which is correct. But worth double checking that conversation_id is still unique across the N copies. If it's generated from scenario_id you could get collisions.

arksim/simulation_engine/simulator.py
```diff
@@ -273,30 +273,31 @@ def on_turn_complete() -> None:
         results: list[ConversationState] = []
 
         for scenario in scenarios.scenarios:
-            if len(running_tasks) >= num_workers:
-                done, running_tasks = await asyncio.wait(
```
itsarbit, line 277: Does the progress bar / total count still work correctly? The total was probably calculated from `len(scenarios)` somewhere upstream, but now the actual conversation count is `len(scenarios) * num_convos_per_scenario`. If the progress callback isn't updated the UI will show wrong totals.

tests/unit/test_simulator_num_convos.py
```diff
@@ -0,0 +1,122 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Tests that Simulator.simulate respects num_convos_per_scenario."""
+
+from __future__ import annotations
+
```
itsarbit, line 5: Nice test coverage. The three cases (1 scenario x N convos, M scenarios x N convos, default 1 convo) cover the important paths well.

arksim/simulation_engine/simulator.py
nrajanee: good point! 

https://github.com/arklexai/arksim/blob/97407f0638ba1c09bd0d062c6a53f9442c938890/arksim/simulation_engine/simulator.py#L247

The estimated_total_turns (used for the progress bar total and the on_progress callback) was already computed as num_convos_per_scenario * len(scenarios) * max_turns. That's also why the progress bar looked wrong before the fix.

arksim/simulation_engine/simulator.py
nrajanee: yup i checked it's unique. thank you!

nrajanee: Each call creates a fresh agent instance on line 130 (agent = create_agent(self.agent_config)), and get_chat_id() returns a new UUID each time — it's not derived from scenario_id. So no collisions. 

https://github.com/arklexai/arksim/blob/97407f0638ba1c09bd0d062c6a53f9442c938890/arksim/simulation_engine/simulator.py#L132

PR #60, opened 2026-03-04
fix: multi-provider chat completions and config-driven LLM

salianahshal2:
## Summary

Adds multi-provider support to the chat completions agent (OpenAI, Anthropic, Gemini), makes example custom metrics use the LLM from `config.yaml`, and fixes the HTML report when there are no conversations.

Closes #

## Changes

- **Chat completions client** (`arksim/simulation_engine/agent/clients/chat_completions.py`): Parse OpenAI-style (`choices`), Anthropic-style (`content` list), and Gemini-style (`candidates`) responses in `_extract_content`; handle HTTP 4xx/5xx and surface API error payloads.
- **Tests**: Add `tests/unit/test_chat_completions.py` for `_extract_content` (OpenAI, Anthropic, Gemini, and unsupported/error cases).
- **Examples – config-driven LLM**: In `examples/bank-insurance/custom_metrics.py` and `examples/e-commerce/custom_metrics.py`, load `model` and `provider` from `config.yaml` in the same directory; require config (no fallbacks).
- **Dependencies**: Pin `anthropic>=0.77.1` and `google-genai>=1.26.0` in optional extras.
- **E-commerce example**: Point `examples/e-commerce/config.yaml` at Gemini (`gemini-2.5-flash`, `provider: gemini`) for both agent and evaluation.
- **HTML report**: In `generate_html_report`, allow empty `convo_rows` instead of raising, so the report generates when simulation has 0 conversations (e.g. all runs failed).

## Documentation

- [x] No docs needed (explain why below)

No docs needed here as its being updated in a different PR.

## How to Test

- [ ] `ruff check .` passes
- [ ] `pytest tests/` passes
- [ ] Manual verification: Run `arksim simulate-evaluate config.yaml` from `examples/e-commerce` with `GEMINI_API_KEY` set (or from `examples/bank-insurance` with `ANTHROPIC_API_KEY`) and confirm simulation/evaluation and HTML report; with 0 conversations, report generation should not raise.

## Notes

- Chat completions agent remains backward compatible: OpenAI-style responses are still parsed first; Anthropic and Gemini are additive.
- E-commerce is configured for Gemini; bank-insurance remains on Claude. Custom metrics in both examples now read `model` and `provider` from the same directory’s `config.yaml`.

## Reviewers

### DIFF
--- a/arksim/utils/html_report/generate_html_report.py
+++ b/arksim/utils/html_report/generate_html_report.py
@@ -494,6 +494,9 @@ def generate_html_report(params: HtmlReportParams) -> Path:
     convo_rows = _build_convo_rows(
         params.evaluation, params.simulation, params.scenarios
     )
+    if not convo_rows:
+        raise ValueError("conversations is required (should be in evaluation)")
+
     turn_rows = _build_turn_rows(params.evaluation)
     error_rows = _build_error_rows(
         params.evaluation, params.simulation, params.chat_id_to_label


arksim/utils/html_report/generate_html_report.py
```diff
@@ -494,9 +494,6 @@ def generate_html_report(params: HtmlReportParams) -> Path:
     convo_rows = _build_convo_rows(
         params.evaluation, params.simulation, params.scenarios
     )
-    if not convo_rows:
```
shehan360, line 497: undo this

PR #63, opened 2026-03-04
feat: support in-memory evaluation flow and refresh docs

shehan360:
## Summary

Enable a direct Python simulation → evaluation workflow by allowing `run_evaluation` to consume in-memory objects, and update docs/examples to reflect the new usage pattern.

Closes #

## Changes

- Added optional in-memory inputs to evaluation flow (`simulation` and `scenarios`) in `run_evaluation`
- Improved validation/error handling when no simulation object is passed and no `simulation_file_path` is provided
- Updated evaluator input handling to support in-memory use without requiring simulation output path at construction time
- Refreshed Python docs/examples in `docs/quickstart.mdx` and `docs/evaluate-conversation.mdx` to show end-to-end in-memory usage
- Added/updated quickstart example script for the new flow
- Deleted `docs/README.md` as it was not used anywhere.

## How to Test

- [ ] `ruff check .` passes
- [ ] `pytest tests/` passes
- [x] Manual verification: ran the quickstart Python flow and confirmed simulation output can be passed directly into `run_evaluation` (with optional scenarios) and HTML/report outputs generate as expected

## Notes

- This keeps file-based evaluation fully supported while enabling a cleaner SDK-first workflow.
- No breaking API changes expected; new params are optional and backward-compatible.

## Reviewers

### DIFF
--- a/tests/unit/test_run_evaluation.py
+++ b/tests/unit/test_run_evaluation.py
@@ -185,10 +185,15 @@ def test_run_evaluation_passes_in_memory_scenarios_to_html_report(
     report_module = importlib.import_module(
         "arksim.utils.html_report.generate_html_report"
     )
+    sentinel_tag = "patched-html-params"
+
+    def fake_html_report_params(**kwargs: object) -> SimpleNamespace:
+        return SimpleNamespace(_sentinel=sentinel_tag, **kwargs)
+
     monkeypatch.setattr(
         report_module,
         "HtmlReportParams",
-        lambda **kwargs: SimpleNamespace(**kwargs),
+        fake_html_report_params,
     )
     monkeypatch.setattr(
         report_module,
@@ -199,4 +204,5 @@ def test_run_evaluation_passes_in_memory_scenarios_to_html_report(
     run_evaluation(settings, simulation=simulation, scenarios=scenarios)
 
     assert len(html_calls) == 1
+    assert getattr(html_calls[0], "_sentinel", None) == sentinel_tag
     assert html_calls[0].scenarios is scenarios
--- a/tests/unit/test_cli.py
+++ b/tests/unit/test_cli.py
@@ -5,6 +5,7 @@ import io
 import os
 import sys
 import tarfile
+from pathlib import Path
 from unittest.mock import MagicMock, patch
 
 import pytest
@@ -15,6 +16,7 @@ from arksim.cli import (
     _parse_value,
     _run_examples,
     build_parser,
+    main,
     parse_extra_args,
     validate_overrides,
 )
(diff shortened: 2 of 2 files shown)


xinyang-articulate (approved):

arksim/evaluator/entities.py
```diff
@@ -76,15 +76,10 @@ class EvaluationInput(BaseModel):
     )
 
     @model_validator(mode="after")
```
itsarbit, line 78: Removing the `simulation_file_path` validation entirely means `arksim evaluate config.yaml` (standalone, no pipeline) silently accepts a missing path and fails later with an unclear error inside `run_evaluation`. The `skip_input_dir_validation` context flag was the right pattern for the pipeline case. Suggest keeping the validator and the context check here, and only skip it in `run_evaluation` when `simulation` is passed in-memory.

arksim/cli.py
```diff
@@ -468,10 +468,7 @@ def main() -> None:
         evaluation_settings = {
             k: v for k, v in settings.items() if k in EvaluationInput.model_fields
         }
-        evaluation_input = EvaluationInput.model_validate(
-            evaluation_settings,
-            context={"skip_input_dir_validation": True},
```
itsarbit, line 473: This change is coupled to the validator removal in entities.py. If the validator is restored (per my other comment), this needs to stay as `EvaluationInput.model_validate(evaluation_settings, context={"skip_input_dir_validation": True})` so the pipeline path does not reject the missing `simulation_file_path`.

tests/unit/test_run_evaluation.py
```diff
@@ -0,0 +1,202 @@
...
+    monkeypatch.setattr(
+        "arksim.scenario.Scenarios.load",
+        lambda _path: (_ for _ in ()).throw(
+            AssertionError("Scenarios.load should not be called")
+        ),
+    )
+
+    class FakeEvaluator:
+        def __init__(self, *args: object, **kwargs: object) -> None:
+            self.chat_id_to_label = {"conv-1": "Conversation 1"}
```
itsarbit, line 170: `monkeypatch.setattr(report_module, ...)` patches the module object, but `run_evaluation` does a lazy `from ... import HtmlReportParams` inside the function body, creating a fresh local binding that bypasses this patch. This test likely passes but does not actually exercise the HTML path correctly. Use `monkeypatch.setattr("arksim.evaluator.evaluator.HtmlReportParams", ...)` instead, or move the import to module level so patching the module works.

docs/README.md
```diff
@@ -1,158 +0,0 @@
-# Arksim
```
itsarbit: Deleting this file is fine if it is stale, but it is not mentioned in the PR description or CHANGELOG. Worth calling out explicitly so reviewers know it is intentional and not an accidental deletion.

arksim/evaluator/entities.py
shehan360: Good call on preserving early failure for standalone evaluate.
I addressed this by moving the strict simulation_file_path checks into the standalone CLI path (arksim evaluate) instead of keeping them in EvaluationInput.

Specifically, in cli.py under args.command == "evaluate" we now fail early with simple errors if the path is missing or does not exist, before run_evaluation is called.

This keeps standalone behavior explicit and user-friendly, while still allowing valid in-memory Python usage (run_evaluation(..., simulation=simulation)) without special model context flags.

run_evaluation still retains its runtime guard as a fallback.

arksim/cli.py
shehan360: Agreed on the coupling concern. With the new approach, the coupling is removed entirely: simulate-evaluate no longer needs model_validate(..., context=...) because EvaluationInput no longer enforces file-path presence at model level.

Pipeline compatibility is preserved naturally (in-memory simulation is passed directly), and standalone strictness is handled where it actually matters: the standalone evaluate CLI branch.

tests/unit/test_run_evaluation.py
shehan360: Done!

docs/README.md
shehan360: Yes this is not used anywhere. Confirmed with salia.

PR #76, opened 2026-03-05
feat!: support custom Python agent connector

shehan360:
## Summary

Previously, connecting an agent to arksim required running it as an HTTP server — either as a chat completions endpoint (`agent_type: chat_completions`) or via the Agent-to-Agent protocol (`agent_type: a2a`). This meant users had to stand up and manage a separate server process even for simple Python agents, adding unnecessary infrastructure overhead.

This PR adds `agent_type: custom`, which loads a Python `BaseAgent` subclass directly at runtime — no HTTP server required. Users point to a `.py` file, arksim dynamically imports it, and the agent runs in-process alongside the simulation. This is especially useful for agents built with frameworks like the OpenAI Agents SDK, LangChain, or any custom Python logic that doesn't need to be wrapped in an API layer.

## Changes

- Add `CUSTOM` variant to `AgentType` enum
- Add `CustomConfig` model (`module_path`, optional `class_name`) and wire into `AgentConfig` under a `config` field
- Add `CustomAgent` wrapper with dynamic module loading and auto-discovery of the `BaseAgent` subclass when `class_name` is omitted
- Add `custom` branch to `create_agent()` factory
- Fail simulation early with `RuntimeError` if no conversations complete (prevents silent pass-through to evaluation)
- Remove `logging.basicConfig()` from example loaders to prevent duplicate log output
- Add custom agent wrappers and `config_custom.yaml` to bank-insurance and e-commerce examples
- Add 21 unit tests covering config validation, factory, execution, auto-discovery, and error handling

## Documentation

- [x] Updated `README.md` (if installation, quickstart, or usage changed)
- [ ] No docs needed (explain why below)

Examples serve as documentation for now; docs update to follow in a separate PR.

## How to Test

- [x] `ruff check .` passes
- [x] `pytest tests/` passes
- [x] Manual verification: run `arksim simulate-evaluate config_custom.yaml` from `examples/bank-insurance/` with a valid OpenAI key

## Notes

- Custom agents must track their own conversation history — the engine only passes the current turn's query via `execute()`
- Auto-discovery scans the module for `BaseAgent` subclasses; errors clearly if zero or multiple are found
- The module's parent directory is added to `sys.path` so sibling package imports work

## Reviewers

### DIFF
--- a/arksim/simulation_engine/agent/clients/custom.py
+++ b/arksim/simulation_engine/agent/clients/custom.py
@@ -105,6 +105,8 @@ class CustomAgent(BaseAgent):
                 self.custom_config.module_path,
                 self.custom_config.class_name,
             )
+        # The inner agent receives the full agent_config (including
+        # agent_type="custom") because BaseAgent.__init__ requires it.
         return cls(self.agent_config)
 
     async def get_chat_id(self) -> str:
--- a/arksim/simulation_engine/entities.py
+++ b/arksim/simulation_engine/entities.py
@@ -90,21 +90,20 @@ class SimulationInput(BaseModel):
             #   - Config-sourced → relative to config file's directory
             #   - CLI-sourced    → relative to cwd (where the CLI ran)
             #   - Already absolute → unchanged
-            if (
-                self.agent_config
-                and self.agent_config.custom_config
-                and hasattr(self.agent_config.custom_config, "module_path")
-                and self.agent_config.custom_config.module_path
-                and not os.path.isabs(self.agent_config.custom_config.module_path)
-            ):
-                module_path = self.agent_config.custom_config.module_path
+            custom_cfg = (
+                getattr(self.agent_config, "custom_config", None)
+                if self.agent_config
+                else None
+            )
+            module_path = (
+                getattr(custom_cfg, "module_path", None) if custom_cfg else None
+            )
+            if module_path and not os.path.isabs(module_path):
                 if "module_path" in cli_overrides:
                     resolved = os.path.abspath(module_path)
                 else:
                     resolved = os.path.normpath(os.path.join(config_dir, module_path))
-                object.__setattr__(
-                    self.agent_config.custom_config, "module_path", resolved
-                )
+                object.__setattr__(custom_cfg, "module_path", resolved)
 
         if not self.agent_config and not self.agent_config_file_path:
             raise ValueError(
--- a/examples/e-commerce/run_pipeline.py
+++ b/examples/e-commerce/run_pipeline.py
@@ -3,6 +3,9 @@
 Example script that runs the full pipeline (simulation -> evaluation)
 programmatically using a custom agent class — no HTTP server required.
 
+For HTTP-based approaches (Chat Completions or A2A), see the corresponding
+config YAML files: config_chat_completions.yaml and config_a2a.yaml.
+
 Usage:
     python run_pipeline.py
 """
--- a/examples/bank-insurance/run_pipeline.py
+++ b/examples/bank-insurance/run_pipeline.py
@@ -3,6 +3,9 @@
 Example script that runs the full pipeline (simulation -> evaluation)
 programmatically using a custom agent class — no HTTP server required.
 
+For HTTP-based approaches (Chat Completions or A2A), see the corresponding
+config YAML files: config_chat_completions.yaml and config_a2a.yaml.
+
 Usage:
     python run_pipeline.py
 """


arksim/simulation_engine/agent/clients/custom.py
```diff
@@ -0,0 +1,147 @@
...
+    spec = importlib.util.spec_from_file_location(module_name, str(path))
+    if spec is None or spec.loader is None:
+        raise ImportError(f"Cannot create module spec from: {path}")
+
+    # Add module's parent dir to sys.path so sibling packages are importable
+    parent_dir = str(path.parent)
+    if parent_dir not in sys.path:
+        sys.path.insert(0, parent_dir)
+
+    module = importlib.util.module_from_spec(spec)
```
itsarbit: `sys.path.insert(0, parent_dir)` is never cleaned up. This permanently mutates the import path for the process. If two custom agents live in different directories but share a sibling module name, the first insertion wins and the second agent imports the wrong module. Same concern with `sys.modules[module_name]` on line 40 - keyed by stem only, so two files named `custom_agent.py` in different directories would collide.

arksim/config/core/agent.py
```diff
@@ -84,13 +84,28 @@ def get_headers(self) -> dict[str, str]:
         return resolved_headers
 
 
+class CustomConfig(BaseModel):
+    """Configuration for custom Python agent type."""
+
```
itsarbit, line 89: Making `api_config` optional (`None` default) weakens the type contract for `chat_completions` and `a2a` agents. The `parse_config` validator catches the missing field, but any code that accesses `config.api_config` after construction now needs a None check. Consider keeping `api_config` required and using a `Union` discriminator, or adding a `model_validator` that asserts `api_config is not None` for non-custom types.

arksim/simulation_engine/agent/clients/custom.py
```diff
@@ -0,0 +1,147 @@
...
+        self.custom_config: CustomConfig = agent_config.config
+        self.chat_id = str(uuid.uuid4())
+        self._inner: BaseAgent | None = None
+
+    def _create_inner_agent(self) -> BaseAgent:
+        """Dynamically load and instantiate the custom agent class."""
+        cls = load_custom_agent_class(
+            self.custom_config.module_path,
+            self.custom_config.class_name,
+        )
```
itsarbit: `_create_inner_agent` passes `self.agent_config` (which has `agent_type: custom`) to the user's class. The inner agent receives config it doesn't use. Consider passing a minimal config or just the fields the inner agent needs.

arksim/simulation_engine/agent/clients/custom.py
```diff
@@ -0,0 +1,147 @@
...
+        return cls(self.agent_config)
+
+    async def get_chat_id(self) -> str:
+        """Get the chat ID."""
+        return self.chat_id
+
+    async def execute(self, user_query: str, **kwargs: object) -> str:
+        """Execute user query by delegating to the custom agent."""
+        if self._inner is None:
```
itsarbit, line 120: This returns `self.chat_id` (generated at init) even after `_inner` is created. The inner agent may have its own meaningful chat ID. Consider delegating: `return await self._inner.get_chat_id() if self._inner else self.chat_id`

examples/bank-insurance/config_custom.yaml
```diff
@@ -0,0 +1,45 @@
+# Custom Agent Connector Configuration
+# Loads the bank-insurance RAG agent directly as a Python class.
+# No HTTP server required — arksim imports the agent module at runtime.
+#
+# Usage: arksim simulate-evaluate config.custom.yaml
```
itsarbit: Typo: `config.custom.yaml` should be `config_custom.yaml` (underscore, not dot).

arksim/simulation_engine/agent/clients/custom.py
```diff
@@ -0,0 +1,147 @@
...
+    - File path: './my_agent.py' or '/abs/path/agent.py'
+    - Dotted module: 'my_package.agent'
+    """
+    if module_path.endswith(".py") or "/" in module_path or "\\" in module_path:
+        module = _load_module_from_path(module_path)
+    else:
+        module = _load_module_from_dotted(module_path)
+
+    if class_name is None:
+        return _find_base_agent_subclass(module)
```
itsarbit, line 70: `module_path` in the config is not resolved relative to the config file directory. `SimulationInput` resolves `scenario_file_path` and `output_file_path` relative to `_config_dir`, but `config.module_path` is left as-is. Running `arksim simulate examples/bank-insurance/config_custom.yaml` from the repo root would fail because `./custom_agent.py` resolves to the current working directory, not the example directory.

arksim/simulation_engine/agent/clients/custom.py
shehan360: `sys.modules` collision is fixed — module names are now UUID-prefixed (`_arksim_{uuid}_{stem}`) so two files with the same stem no longer collide.

`sys.path` entries are guarded against duplicates but intentionally kept after loading — removing them would break deferred imports inside agent methods (e.g. a lazy `from sibling_pkg.utils import helper` inside `execute()`). Updated the comment to document why the entry is retained.

arksim/config/core/agent.py
shehan360: Added a `validate_type_config` after-validator that asserts `api_config is not None` for `chat_completions`/`a2a` types. This catches both dict-based construction (via the before-validator) and programmatic construction (via the after-validator), providing defense-in-depth.

arksim/simulation_engine/agent/clients/custom.py
shehan360: This is a consequence of the `BaseAgent` contract — `__init__` requires an `AgentConfig`. The inner agent stores it via `super().__init__()` but does not inspect `agent_type` or `config`. Leaving as-is since stripping the fields (via `model_construct`) adds complexity for no functional benefit.

arksim/simulation_engine/agent/clients/custom.py
shehan360: Fixed — `get_chat_id()` now delegates to `self._inner.get_chat_id()` when the inner agent exists.

examples/bank-insurance/config_custom.yaml
shehan360: Fixed.

arksim/simulation_engine/agent/clients/custom.py
shehan360: Fixed — `SimulationInput.validate_simulation_input` now resolves `module_path` relative to the config file directory (config-sourced) or cwd (CLI-sourced).

arksim/utils/module_loader.py
```diff
@@ -0,0 +1,69 @@
...
+    if path.suffix != ".py":
+        raise ValueError(f"Module path must be a .py file, got: {path}")
+    if not path.exists():
+        raise FileNotFoundError(f"Module file not found: {path}")
+
+    # Use a unique name to avoid collisions in sys.modules when multiple
+    # files share the same stem (e.g. two different custom_agent.py files).
+    module_name = f"_arksim_{uuid.uuid4().hex[:8]}_{path.stem}"
+
+    spec = importlib.util.spec_from_file_location(module_name, str(path))
```
itsarbit, line 48: Good call using a UUID prefix here - this properly avoids the collision issue I flagged. The `_arksim_` prefix also makes it easy to identify arksim-loaded modules when debugging `sys.modules`.

arksim/utils/module_loader.py
```diff
@@ -0,0 +1,69 @@
...
+    # Add parent directory to sys.path so the loaded module can import
+    # sibling packages (e.g. `from agent_server.core import ...` or
+    # `from my_metrics_helpers import ...`).  The entry is kept after
+    # loading to support deferred imports inside the agent (e.g. lazy
+    # imports in method bodies).  The guard prevents duplicate entries.
+    parent_dir = str(path.parent)
+    if parent_dir not in sys.path:
+        sys.path.insert(0, parent_dir)
+
+    module = importlib.util.module_from_spec(spec)
```
itsarbit, line 61: Makes sense to keep the `sys.path` entry for deferred imports. The comment explaining *why* it's retained is helpful.

One edge case to keep in mind for later: if someone loads two custom agents from different directories that each have a `utils.py`, the first directory wins since it's inserted at position 0. Not a blocker right now but worth a TODO if we ever see bug reports around this.

arksim/config/core/agent.py
```diff
@@ -112,6 +157,28 @@ def parse_config(cls, data: object) -> object:
```
itsarbit, line 157: This after-validator is a clean way to handle it. Catches both the dict path and the programmatic path without duplicating logic.

arksim/simulation_engine/entities.py
```diff
@@ -76,11 +81,29 @@ def validate_simulation_input(self, info: ValidationInfo) -> Self:
...
+            #   - Config-sourced → relative to config file's directory
+            #   - CLI-sourced    → relative to cwd (where the CLI ran)
+            #   - Already absolute → unchanged
+            if (
+                self.agent_config
+                and self.agent_config.config
+                and hasattr(self.agent_config.config, "module_path")
+                and self.agent_config.config.module_path
+                and not os.path.isabs(self.agent_config.config.module_path)
+            ):
```
itsarbit: This block feels a bit verbose with the nested `hasattr` + `config` checks. Could be simplified with something like:

```python
custom_cfg = getattr(self.agent_config, "config", None) if self.agent_config else None
if custom_cfg and getattr(custom_cfg, "module_path", None) and not os.path.isabs(custom_cfg.module_path):
```

Not a blocker - just a readability thought.

examples/e-commerce/run_pipeline.py
```diff
@@ -1,7 +1,7 @@
 # SPDX-License-Identifier: Apache-2.0
 """
-Example script that runs the full pipeline (simulation → evaluation)
-programmatically, with custom metrics passed to the evaluator.
+Example script that runs the full pipeline (simulation -> evaluation)
+programmatically using a custom agent class — no HTTP server required.
 
 Usage:
     python run_pipeline.py
```
itsarbit, line 10: Worth noting: this used to demonstrate the `chat_completions` API path programmatically. Now both examples only show `custom` in their Python scripts. Maybe keep the old version as `run_pipeline_api.py` or add a comment pointing users to the YAML configs for the HTTP-based approach? Not a blocker for this PR.

arksim/simulation_engine/agent/clients/custom.py
```diff
@@ -0,0 +1,126 @@
...
+
+    return cls
+
+
+class CustomAgent(BaseAgent):
+    """Wrapper that loads and delegates to a user-provided BaseAgent subclass."""
+
+    def __init__(self, agent_config: AgentConfig) -> None:
+        super().__init__(agent_config)
+        if agent_config.agent_type != AgentType.CUSTOM.value:
```
itsarbit, line 92: The `_create_inner_agent` still passes the full `agent_config` (with `agent_type: custom`) to the user's class. Your rationale makes sense (BaseAgent contract requires it), but it might trip up users who inspect `self.agent_config.agent_type` inside their custom agent and see `custom` instead of something meaningful to them. A one-line comment here saying the inner agent receives the outer config could save someone debugging time.

arksim/config/core/agent.py
```diff
@@ -84,13 +84,43 @@ def get_headers(self) -> dict[str, str]:
...
     """Agent configuration."""
 
     agent_name: str = Field(..., description="Unique identifier for the agent")
     agent_type: str = Field(..., description="Agent type identifier")
-    api_config: ChatCompletionsConfig | A2AConfig = Field(
-        ..., description="Agent configuration"
+    api_config: ChatCompletionsConfig | A2AConfig | None = Field(
+        None, description="API configuration for chat_completions or a2a agents"
+    )
+    config: CustomConfig | None = Field(
```
xinyang-articulate: Shall we have a better naming for this because it is not a commonly shared config, and will only used by customized agent?

arksim/utils/module_loader.py
shehan360: Agreed — noted for later. Added a comment explaining why the entry is retained (deferred imports). If collisions surface we can namespace by full path.

arksim/simulation_engine/entities.py
shehan360: Good suggestion — simplified with `getattr`-based shorthand. Down from a 6-line nested `and` chain to 3 lines.

examples/e-commerce/run_pipeline.py
shehan360: Added a note in both `run_pipeline.py` docstrings pointing to `config_chat_completions.yaml` and `config_a2a.yaml` for the HTTP-based approaches.

arksim/simulation_engine/agent/clients/custom.py
shehan360: Added a comment above the return explaining the inner agent receives the full `agent_config` because `BaseAgent.__init__` requires it.

arksim/config/core/agent.py
shehan360: Renamed to `custom_config` — parallel to `api_config`. Updated across all source, tests, YAML configs, and docs.

PR #78, opened 2026-03-06
fix!: fail early if error loading custom metrics or file not found

KathyLiu20:
## Summary

Fixes silent failure when loading custom metrics files. Previously, errors (file not found, import error, uninstantiable class) were logged and skipped, allowing evaluation to continue without the user's custom metrics. Now these conditions raise immediately so the problem is surfaced before evaluation runs.

## Changes

- `_load_custom_metrics`: raise `FileNotFoundError` instead of warning + `continue` when a custom metrics file path does not exist
- `_load_custom_metrics`: raise `RuntimeError` (chained from original exception) instead of logging + `continue` when a module fails to load or a metric class cannot be instantiated
- Improved error messages: "Could not load" → "Could not instantiate" to be more precise about what failed

## Documentation

- [x] No docs needed — internal error-handling behavior, no API or config change

## How to Test

- [x] `ruff check .` passes
- [x] `pytest tests/` passes
- [x] Manual verification: pass a nonexistent custom metrics path and confirm the run aborts immediately with a clear error; pass a file with a broken metric class and confirm the same

## Notes

This is a breaking change in behavior for users who relied on invalid custom metrics being silently skipped. The new behavior is strictly safer — it prevents running an evaluation that is missing metrics the user explicitly configured.

## Reviewers

### DIFF
--- a/tests/unit/test_html_report_integration.py
+++ b/tests/unit/test_html_report_integration.py
@@ -4,6 +4,10 @@
 Uses importlib to load modules directly from file paths, bypassing
 the arksim.__init__.py chains that pull in heavy deps
 (langchain, azure, etc.) that may not be installed in the test env.
+
+All sys.modules patching is done inside a module-scoped fixture so that
+monkeypatch restores the original entries automatically after the test
+module finishes — no manual bookkeeping required.
 """
 
 from __future__ import annotations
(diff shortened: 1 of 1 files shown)


shehan360 (changes requested):
## Last commit (`fix tests module load issue`) needs cleanup

A few issues with the `sys.modules` tracking approach in `test_html_report_integration.py`:

1. **`_MODULES_LOADED` is unused** — it gets populated but never read. Remove it.

2. **Module-level mutable globals for test bookkeeping is fragile** — `_MODULES_TO_RESTORE` and `_MODULES_LOADED` as module-level state is what `pytest` fixtures and `monkeypatch` exist for.

3. **Patches the symptom, not the cause** — the real problem is `sys.modules` manipulation at module load time instead of in proper fixtures. This just makes the cleanup more thorough without fixing the underlying approach.

PR #96, opened 2026-03-14
feat: add MiniMax LLM provider support

octo-patch:
## Summary

Add [MiniMax](https://www.minimaxi.com) as a new LLM provider for ArkSim. MiniMax offers high-performance language models through an OpenAI-compatible API, making it straightforward to integrate.

## Changes

- **New provider**: `arksim/llms/chat/providers/minimax.py` - MiniMax LLM provider using OpenAI-compatible Chat Completions API
- **Factory registration**: Added `minimax` provider option in `LLM._get_provider()`
- **Unit tests**: Full test coverage in `tests/unit/test_minimax_llm.py` and factory test in `test_llm_factory.py`
- **README**: Updated provider documentation to include MiniMax

## Supported Models

| Model | Description |
|-------|-------------|
| `MiniMax-M2.5` | Peak performance, 204K context window |
| `MiniMax-M2.5-highspeed` | Same performance, faster and more agile |

## Usage

```bash
export MINIMAX_API_KEY="your-key"
```

```yaml
provider: minimax
model: MiniMax-M2.5
```

```python
from arksim.llms.chat import LLM
llm = LLM(model="MiniMax-M2.5", provider="minimax")
result = llm.call("Hello\!")
```

## Implementation Details

- Uses the bundled OpenAI SDK (no new dependencies) with MiniMax OpenAI-compatible endpoint (`https://api.minimax.io/v1`)
- Handles temperature constraint `(0.0, 1.0]` with automatic clamping
- Structured output via prompt engineering + JSON parsing
- Supports custom base URL via `MINIMAX_BASE_URL` env var

## API Documentation

- [MiniMax OpenAI-Compatible API](https://platform.minimax.io/docs/api-reference/text-openai-api)

## Test Results

- All 26 new unit tests pass
- All 510 existing tests pass (0 regressions)
- Integration tested with MiniMax API (both MiniMax-M2.5 and MiniMax-M2.5-highspeed)

### DIFF
--- a/arksim/llms/chat/providers/minimax.py
+++ b/arksim/llms/chat/providers/minimax.py
@@ -66,9 +66,11 @@ class MiniMaxLLM(BaseLLM):
             "messages": chat_messages,
         }
 
-        # MiniMax requires temperature in (0.0, 1.0]; default to 1.0.
-        temp = self.temperature if self.temperature is not None else 1.0
-        params["temperature"] = max(0.01, min(temp, 1.0))
+        if self.temperature is not None:
+            params["temperature"] = max(0.01, min(self.temperature, 1.0))
+
+        if schema:
+            params["response_format"] = {"type": "json_object"}
 
         return params
 
--- a/arksim/llms/chat/providers/minimax.py
+++ b/arksim/llms/chat/providers/minimax.py
@@ -2,6 +2,7 @@
 from __future__ import annotations
 
 import json
+import logging
 import os
 import re
 from typing import Any, TypeVar, overload
@@ -14,6 +15,8 @@ from arksim.llms.chat.base.types import LLMMessage
 from arksim.llms.chat.utils import retry
 
 T = TypeVar("T", bound=BaseModel)
+
+logger = logging.getLogger(__name__)
 
 DEFAULT_BASE_URL = "https://api.minimax.io/v1"
 
@@ -67,7 +70,15 @@ class MiniMaxLLM(BaseLLM):
         }
 
         if self.temperature is not None:
-            params["temperature"] = max(0.01, min(self.temperature, 1.0))
+            clamped = max(0.01, min(self.temperature, 1.0))
+            if clamped != self.temperature:
+                logger.warning(
+                    "MiniMax requires temperature in (0.0, 1.0]; "
+                    "clamping %.2f to %.2f",
+                    self.temperature,
+                    clamped,
+                )
+            params["temperature"] = clamped
 
         if schema:
             params["response_format"] = {"type": "json_object"}
@@ -111,6 +122,8 @@ class MiniMaxLLM(BaseLLM):
         params = self._prepare_params(messages, schema=schema)
         response = self.client.chat.completions.create(**params)
         content = response.choices[0].message.content
+        if content is None:
+            raise ValueError("MiniMax returned empty response")
         if schema:
             parsed = self._parse_json(content)
             return schema.model_validate(parsed)
@@ -136,6 +149,8 @@ class MiniMaxLLM(BaseLLM):
         params = self._prepare_params(messages, schema=schema)
         response = await self.async_client.chat.completions.create(**params)
         content = response.choices[0].message.content
+        if content is None:
+            raise ValueError("MiniMax returned empty response")
         if schema:
             parsed = self._parse_json(content)
             return schema.model_validate(parsed)
--- a/arksim/llms/chat/providers/minimax.py
+++ b/arksim/llms/chat/providers/minimax.py
@@ -1,0 +1,140 @@
+# SPDX-License-Identifier: Apache-2.0
+from __future__ import annotations
+
+import json
+import os
+import re
+from typing import Any, TypeVar, overload
+
+from openai import AsyncOpenAI, OpenAI
+from pydantic import BaseModel
+
+from arksim.llms.chat.base.base_llm import BaseLLM
+from arksim.llms.chat.base.types import LLMMessage
+from arksim.llms.chat.utils import retry
+
+T = TypeVar("T", bound=BaseModel)
+
+DEFAULT_BASE_URL = "https://api.minimax.io/v1"
+
+
+class MiniMaxLLM(BaseLLM):
+    def __init__(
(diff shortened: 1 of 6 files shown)


arksim/llms/chat/providers/minimax.py
```diff
@@ -0,0 +1,140 @@
...
+        chat_messages = self._prepare_messages(messages)
+
+        if schema:
+            schema_json = json.dumps(schema.model_json_schema())
+            system_prompt = (
+                "You must respond with valid JSON only, no extra text. "
+                f"JSON Schema: {schema_json}"
+            )
+            chat_messages.insert(0, {"role": "system", "content": system_prompt})
+
```
itsarbit, line 66: other providers use native structured output (OpenAI `text_format`, Anthropic `output_format`). does MiniMax's endpoint support `response_format`? that'd be way more reliable than prompt engineering + regex parsing. if it doesn't then this approach is totally fine, just want to make sure we tried the native path first.

arksim/llms/chat/providers/minimax.py
```diff
@@ -0,0 +1,140 @@
...
+        params: dict[str, Any] = {
+            "model": self.model,
+            "messages": chat_messages,
+        }
+
+        # MiniMax requires temperature in (0.0, 1.0]; default to 1.0.
+        temp = self.temperature if self.temperature is not None else 1.0
+        params["temperature"] = max(0.01, min(temp, 1.0))
+
```
itsarbit: other providers skip temperature when the user hasn't set one and let the API use its own default. always sending 1.0 means more random outputs than users might expect. is there a reason MiniMax needs an explicit value here?

itsarbit (commented):
hey, welcome and thanks for the contribution! clean implementation and thorough test coverage.

left a couple inline questions. nothing major, just want to make sure we're aligned on structured output and temperature defaults before merging.

arksim/llms/chat/providers/minimax.py
octo-patch: Good call! MiniMax does support `response_format` via their OpenAI-compatible API. Updated to use native `response_format: {"type": "json_object"}` when a schema is requested. The system prompt with the JSON schema is kept as well to guide the model toward the correct structure, but the API-level constraint ensures valid JSON output. Fixed in c152382.

arksim/llms/chat/providers/minimax.py
octo-patch: You're right — removed the hardcoded temperature default. Now it only sends temperature when explicitly set by the user, matching the behavior of the OpenAI and Anthropic providers. The clamping to (0.01, 1.0] is still applied when a value is provided, since MiniMax rejects values outside that range. Fixed in c152382.

itsarbit (commented):
Thanks for the contribution! This is a really clean integration - love that it reuses the OpenAI SDK so there's no new dependency to pull in. The test coverage is solid too.

Left a couple of inline comments, one is a minor bug the rest are just small suggestions. Nice work overall!

One other thing - looks like the supported models table in the PR description still references M2.5, but the code was updated to default to M2.7 in a follow-up commit. Might want to update the description so it matches what's in `app.js`.

arksim/llms/chat/providers/minimax.py
```diff
@@ -0,0 +1,142 @@
...
+    def call(
+        self, messages: str | list[LLMMessage], schema: None = None, **kwargs: object
+    ) -> str: ...
+
+    @retry()
+    def call(
+        self,
+        messages: str | list[LLMMessage],
+        schema: type[T] | None = None,
+        **kwargs: object,
```
itsarbit, line 120: Small thing - the OpenAI SDK can return `None` here if the model response has no content (rare but possible). Might be worth adding a quick guard so it doesn't surface as a confusing downstream error:

```python
if content is None:
    raise ValueError("MiniMax returned empty response")
```

Same applies to the async version below.

arksim/llms/chat/providers/minimax.py
```diff
@@ -0,0 +1,142 @@
...
+            params["temperature"] = max(0.01, min(self.temperature, 1.0))
+
+        if schema:
+            params["response_format"] = {"type": "json_object"}
+
+        return params
+
+    @staticmethod
+    def _parse_json(text: str) -> dict[str, Any]:
+        """Extract JSON from the response text.
```
itsarbit, line 90: Totally makes sense given MiniMax's constraint. One thought - if someone explicitly passes `temperature=0.0` expecting deterministic output, the silent bump to 0.01 could be a bit surprising. Maybe worth a `logger.warning` so they know? Totally optional though.

arksim/llms/chat/providers/minimax.py
octo-patch: Good catch! Added the null guard for both `call()` and `call_async()` — raises `ValueError("MiniMax returned empty response")` if content is `None`. Also added test cases for both sync and async paths. Fixed in 4c2c018.

arksim/llms/chat/providers/minimax.py
octo-patch: Great suggestion — added a `logger.warning` that fires whenever the temperature gets clamped. This way users passing `temperature=0.0` will see a clear message explaining that MiniMax requires values in `(0.0, 1.0]` and their value was adjusted to `0.01`. Added tests to verify the warning is emitted (and not emitted for valid values). Fixed in 4c2c018.

itsarbit (approved):
LGTM! Clean integration, solid test coverage, and all feedback addressed.

nit: PR description still references M2.5 in the supported models table but the code defaults to M2.7 now. Worth updating for anyone reading this later.

itsarbit (approved):

octo-patch:
Thank you for the thorough review and approval @itsarbit! Happy to make any further adjustments if needed.

itsarbit:
Hey @octo-patch, this has a merge conflict in README.md now since we restructured the README on main. Could you rebase against main and resolve the conflict? The code files should merge cleanly. The only conflict is in README.md where we simplified it to link to docs instead of inlining everything. Just take main's version of README.md and it should be good.

PR #129, opened 2026-04-01
feat: add LLM token usage tracking to simulation and evaluation

xinyang-articulate:
## Summary

Add automatic token usage tracking across all LLM providers (OpenAI, Anthropic, Google) using context variables. Usage totals are attached to simulation and evaluation outputs and displayed in the HTML report.

## Changes

- New `arksim/llms/chat/base/usage.py` module: `TokenUsage` (Pydantic model exported as a public API), `UsageRecord`, `UsageTracker`, plus `usage_run` / `usage_label` / `usage_tags` context managers and
a `track_usage` helper backed by `ContextVar`.
- `UsageRecord` carries typed context dimensions (`module`, `run_id`, `component`, `step`, `conversation_id`, `turn_id`) and counts for input, output, cache read, cache creation, reasoning, and total
tokens, with raw `input_token_details` / `output_token_details` passthrough.
- Anthropic, Google, and OpenAI adapters call `track_usage` after each sync/async response, normalizing provider-specific fields (Anthropic cache read/create, Google `usage_metadata`, OpenAI
`input_tokens_details` / reasoning tokens) via `clean_usage_value`.                                                                                                                                       
- `Simulator.simulate_conversation` and `decide_ask_or_answer` wrap their LLM calls in `usage_label` so per-conversation and multi-knowledge usage are tagged correctly.
- `run_simulation` and `run_evaluation` now open a `usage_run` scope, attach the resulting `TokenUsage` to `Simulation` / `Evaluation`, and emit a summary log line. `Evaluator` propagates the active    
context with `contextvars.copy_context().run` when dispatching turn/conversation work to thread pools.                                                                                                    
- `Simulation` and `Evaluation` Pydantic models gain an optional `usage: TokenUsage | None` field (defaults to `None`, preserving backward compatibility with older saved JSON).                          
- HTML report: new `TokenUsageSummary` data class on `ReportSummary`, plus a "Token Usage" section in `report_template.html` rendering an aggregated per-model table and collapsible breakdowns (by       
conversation for simulation, by phase for evaluation).                                                                                                                                                    
- Public API: `TokenUsage` re-exported from `arksim` via the lazy-import map. 

## Documentation

## How to Test

- [x] `ruff check .` passes
- [x] `pytest tests/` passes
- [x] Manual verification: run a simulation and evaluation, verify `usage` field is populated in the output JSON and the HTML report shows a "Token Usage" section

## Notes

- `TokenUsage` fields are optional (`None`) so existing serialized outputs remain compatible
- `contextvars.copy_context().run` is used to propagate the tracker into thread pool workers for turn evaluation and goal completion scoring

## Reviewers


arksim/evaluator/entities.py
```diff
@@ -228,6 +228,14 @@ class ConversationEvaluation(BaseModel):
     turn_scores: list[TurnEvaluation]
 
 
+class TokenUsage(BaseModel):
```
KathyLiu20: **Duplication: `TokenUsage` is defined three times.**

`TokenUsage` is defined identically here and in `arksim/simulation_engine/entities.py`, with a near-identical `TokenUsageSummary` in `generate_html_report.py`. Consider defining it once in `arksim/llms/chat/base/usage.py` (which already owns this concept) and importing it everywhere else.

arksim/llms/chat/base/usage.py
```diff
@@ -0,0 +1,122 @@
...
+_current_tracker: ContextVar[UsageTracker | None] = ContextVar(
+    "llm_usage_tracker", default=None
+)
+
+
+def get_current_tracker() -> UsageTracker | None:
+    return _current_tracker.get()
+
+
+def set_current_tracker(tracker: UsageTracker) -> object:
```
KathyLiu20: **Type annotation loss on `set_current_tracker` / `reset_current_tracker`.**

Returning `object` loses type information and forces the `# type: ignore[arg-type]` on line 110. Use `Token` from `contextvars` instead:

```python
from contextvars import Token

def set_current_tracker(tracker: UsageTracker) -> Token[UsageTracker | None]:
    return _current_tracker.set(tracker)

def reset_current_tracker(token: Token[UsageTracker | None]) -> None:
    _current_tracker.reset(token)
```

arksim/llms/chat/base/usage.py
```diff
@@ -0,0 +1,122 @@
...
+        output_tokens: int,
+    ) -> None:
+        self.records.append(
+            UsageRecord(
+                model=model,
+                provider=provider,
+                input_tokens=input_tokens,
+                output_tokens=output_tokens,
+            )
+        )
```
KathyLiu20: **Thread safety of `records.append()`.**

Multiple threads in the `ThreadPoolExecutor` call `record()` concurrently. `list.append()` is GIL-protected in CPython and happens to be safe, but this is an implicit reliance on an implementation detail. A `threading.Lock` around the append would make the intent explicit and be safe under any Python implementation:

```python
import threading

@dataclass
class UsageTracker:
    records: list[UsageRecord] = field(default_factory=list)
    _lock: threading.Lock = field(default_factory=threading.Lock, repr=False, compare=False)

    def record(self, ...) -> None:
        with self._lock:
            self.records.append(...)
```

arksim/llms/chat/providers/openai.py
```diff
@@ -68,6 +69,13 @@ def call(
     ) -> T | str:
         params = self._prepare_params(messages, schema=schema)
         response = self.client.responses.parse(**params)
+        if response.usage:
```
KathyLiu20: **Inconsistency with Anthropic and Google providers.**

The Anthropic and Google providers extract tracking into a `_track_response_usage` helper method, but OpenAI inlines the same logic twice (here and in `call_async`). For consistency, extract a `_track_response_usage` method here as well.

arksim/llms/chat/base/usage.py
```diff
@@ -0,0 +1,122 @@
...
+        return sum(r.input_tokens for r in self.records)
+
+    @property
+    def total_output_tokens(self) -> int:
+        return sum(r.output_tokens for r in self.records)
+
+    def summary(self) -> dict[str, dict[str, int]]:
+        """Per-model breakdown: {model: {input_tokens, output_tokens}}."""
+        breakdown: dict[str, dict[str, int]] = {}
+        for r in self.records:
```
KathyLiu20: **`by_model` silently merges tokens across providers.**

The breakdown is keyed by model name only. If two providers ever use a model with the same name, their tokens would be merged silently. The `provider` field on `UsageRecord` is already collected but unused here. Consider keying by `f"{r.provider}/{r.model}"` to avoid collisions.

itsarbit (changes requested):
Good feature, the token tracking concept is something we need. A few structural things to sort out:

1. TokenUsage is defined in three places (simulation entities, evaluator entities, HTML report module). These will drift. Let's have one canonical model in a shared location and import it everywhere else.

2. The contextvars propagation into ThreadPoolExecutor is the trickiest part of this PR and there are no tests covering it. Can you add at least one test that verifies tracking works correctly across thread boundaries?

3. by_model is typed as dict[str, dict[str, int]] with string keys "input_tokens"/"output_tokens". When we add cache tokens or reasoning tokens down the road, there's no schema to catch missed sites. A small TypedDict or dataclass would make this extensible safely.

4. The providers are inconsistent in how they extract usage. OpenAI uses an inline if-guard, Anthropic uses a helper method, Google uses double getattr with or-0 fallback. Pick one pattern and apply it across all three so future providers have a clear template to follow.

5. The test checkbox in the PR description is unchecked. Please make sure tests pass before requesting another review.

Happy to re-review once these are addressed.

arksim/evaluator/evaluator.py
```diff
@@ -276,6 +283,7 @@ def _on_turn_complete() -> None:
                 inner_workers = 0 if self.params.num_workers == "auto" else num_workers
                 turn_futures = {
                     executor.submit(
+                        contextvars.copy_context().run,
                         evaluate_turn,
```
kmchen-arklex, line 302: contextvars.copy_context().run(...) wraps only the first argument, not the full call — evaluate_turn never sees the tracker


```suggestion
                        ctx = contextvars.copy_context()
turn_futures = {
    executor.submit(ctx.run, evaluate_turn, self.llm, turn_item, ...): turn_item
    for turn_item in turn_items
}
```

arksim/evaluator/evaluator.py
```diff
@@ -318,6 +326,7 @@ def _on_turn_complete() -> None:
         with ThreadPoolExecutor(max_workers=gc_max_workers) as executor:
             gc_futures = {
                 executor.submit(
+                    contextvars.copy_context().run,
```
kmchen-arklex, line 345: ditto

arksim/llms/chat/base/usage.py
kmchen-arklex:
```suggestion
        def summary(self) -> dict[str, dict[str, int]]:
    """Per-model breakdown: {provider/model: {input_tokens, output_tokens}}."""
    breakdown: dict[str, dict[str, int]] = {}
    for r in self.records:
        key = f"{r.provider}/{r.model}"
        if key not in breakdown:
            breakdown[key] = {"input_tokens": 0, "output_tokens": 0}
        breakdown[key]["input_tokens"] += r.input_tokens
        breakdown[key]["output_tokens"] += r.output_tokens
    return breakdown
```

arksim/evaluator/evaluator.py
```diff
@@ -931,7 +940,23 @@ def run_evaluation(
...
+        reset_current_tracker(token)
+
+    evaluator_output.usage = TokenUsage(
+        total_input_tokens=tracker.total_input_tokens,
+        total_output_tokens=tracker.total_output_tokens,
+        by_model=tracker.summary(),
+    )
+    tracker.log_summary()
+
+    evaluator.display_evaluation_summary()
```
itsarbit: **Duplicate `display_evaluation_summary()` call.** Line 958 (new) and line 961 (pre-existing) both call `evaluator.display_evaluation_summary()`. The summary prints twice in every evaluation run. Almost certainly a copy-paste artifact from inserting the tracker block above. Drop line 958.

arksim/llms/chat/base/usage.py
```diff
@@ -0,0 +1,135 @@
...
+                    model=model,
+                    provider=provider,
+                    input_tokens=input_tokens,
+                    output_tokens=output_tokens,
+                )
+            )
+
+    @property
+    def total_input_tokens(self) -> int:
+        return sum(r.input_tokens for r in self.records)
```
itsarbit: **Unlocked `sum()` races list append.** `total_input_tokens` and `total_output_tokens` (line 81) iterate `self.records` without taking `_lock`, while `record()` appends under lock. Concurrent read-while-append is undefined per the language spec; observed `RuntimeError: list changed size during iteration` under free-threaded CPython 3.13+ and during GC. `summary()` at line 83 has the same issue.

Fix: snapshot under the lock, e.g.
```python
@property
def total_input_tokens(self) -> int:
    with self._lock:
        records = list(self.records)
    return sum(r.input_tokens for r in records)
```

arksim/llms/chat/base/usage.py
```diff
@@ -0,0 +1,135 @@
...
+
+
+@dataclass
+class UsageRecord:
+    """Token usage from a single LLM call."""
+
+    model: str
+    provider: str
+    input_tokens: int
+    output_tokens: int
```
itsarbit: **`UsageRecord` missing attribution fields.** With only `{model, provider, input_tokens, output_tokens}`, a practitioner cannot answer "which scenario burned the most tokens?" or "did my LLM-judge cost 3x the agent?" from the records list — which is the main reason eval tools add usage tracking.

Minimum viable additions: `turn_id: str | None`, `scenario_id: str | None`, `conversation_id: str | None`, `call_purpose: Literal["simulated_user", "agent", "judge", "other"] | None`, `timestamp: datetime`. Every competitor (LangFuse `observation.usage`, Braintrust step-level, Maxim session/trace/span, Helicone per-request) attributes at span level.

Also worth splitting `input_tokens` into `prompt_tokens`, `cached_input_tokens`, and `cache_creation_tokens`, and exposing `reasoning_tokens` separately — see the provider-level comments for the per-SDK sources.

arksim/llms/chat/base/usage.py
```diff
@@ -0,0 +1,135 @@
...
+    """Reset the tracker context variable using the token from set_current_tracker."""
+    _current_tracker.reset(token)
+
+
+def track_usage(
+    model: str,
+    provider: str,
+    input_tokens: int,
+    output_tokens: int,
+) -> None:
```
itsarbit, line 334: **No validation + no context-manager wrapper.** Two issues on this function:

1. `track_usage` accepts `input_tokens` / `output_tokens` verbatim. Some SDKs report `-1` when usage is unknown; accumulating that into totals gives negative or garbage numbers. Clamp at the boundary: `tokens = max(0, int(tokens))`.
2. The docstring example at the top of `UsageTracker` shows `set_current_tracker` / `reset_current_tracker` usage manually. A `@contextmanager` wrapper like `with usage_scope() as tracker: ...` would (a) eliminate the "forgot the `finally`" class of scope-leak bugs and (b) be the shape the docstring already implies.

Suggested:
```python
from contextlib import contextmanager

@contextmanager
def usage_scope() -> Iterator[UsageTracker]:
    tracker = UsageTracker()
    token = set_current_tracker(tracker)
    try:
        yield tracker
    finally:
        reset_current_tracker(token)
```

Then `simulator.py` and `evaluator.py` both collapse to `with usage_scope() as tracker: ...`.

itsarbit (commented):
Reviewed via the five-perspective pattern (architect / first-timer / adversary / domain-expert / future-maintainer). Inline comments below cover specific `file:line` items. Five threads that do not pin to a single line:

1. **Docs gap.** PR checks "no docs needed" but this adds two new top-level public exports, a new field on two public output entities (`Simulation.usage`, `Evaluation.usage`), and a new HTML report section. Mintlify docs at `docs/` are untouched. A short page in docs covering `result.usage` access + token tracking scope semantics would close this.

2. **Competitive depth.** LangFuse / Helicone / Braintrust / Maxim all ship per-span attribution, cache-token breakout, reasoning-token breakout, and $ cost. This PR lands at run-level counters with no per-scenario/per-turn/per-call attribution, cache tokens silently collapsed into `input_tokens` (3-10x over-count when caching is on), no reasoning tokens, and no cost mapping. Plumbing is solid; the data surface is 2023-era. Worth a follow-up plan on these four axes.

3. **Provider instrumentation duplication.** `_track_response_usage` is copy-pasted three times across `openai.py`, `anthropic.py`, `google.py`, and already diverging (only Google coerces `None` → `0`). Extract a `BaseLLM._record_usage(response)` hook or an `extract_usage(response) -> (input, output, model, provider) | None` strategy keyed per-provider. The next provider addition will have to guess the contract otherwise.

4. **Parallel telemetry channel.** The repo already has `arksim.tracing` (trace receiver wired at `simulator.py:506`). This PR adds a separate usage-tracking stack rather than emitting tokens as trace events. Document the split, or plan to unify — a future maintainer will ask "why two systems?"

5. **Nested executor coverage.** Any nested `ThreadPoolExecutor` or library-spawned thread in the evaluation path that is NOT wrapped in `copy_context().run(...)` will silently lose the tracker and under-count. The tests verify the ContextVar pattern works in isolation but do not assert "every evaluator call site the simulator uses is wrapped." Worth a regression test that runs a real `evaluate_turn` and asserts per-metric tokens are tracked.

Not marking as request-changes. Flagging so you can prioritize.

arksim/llms/chat/providers/anthropic.py
```diff
@@ -72,6 +73,16 @@ def _prepare_params(
...
         return params
 
+    def _track_response_usage(self, response: object) -> None:
+        usage = getattr(response, "usage", None)
+        if usage:
+            track_usage(
+                self.model,
+                "anthropic",
+                usage.input_tokens,
+                usage.output_tokens,
```
itsarbit: **`None`-attribute crash + prompt-cache tokens silently collapsed.** Two problems on these three lines:

1. `usage.input_tokens` / `usage.output_tokens` are accessed as raw attributes. Anthropic returns `None` for these fields on some cache-heavy and retry responses. `tracker.record(..., None, None)` stores `None`, then `sum(r.input_tokens for r in records)` raises `TypeError` at the very end of a multi-hour run. Only the Google path (`google.py:86`) defensively coerces via `getattr(..., 0) or 0`.
2. Anthropic also returns `cache_creation_input_tokens` and `cache_read_input_tokens` separately. Both are currently thrown away. A practitioner with prompt caching on (standard in 2026) will see reported input tokens 3-10x over their actual billed input. LangFuse/Helicone/Braintrust all break cache out.

Suggested:
```python
usage = getattr(response, "usage", None)
if usage is None:
    return
track_usage(
    model=self.model,
    provider="anthropic",
    input_tokens=max(0, int(getattr(usage, "input_tokens", 0) or 0)),
    output_tokens=max(0, int(getattr(usage, "output_tokens", 0) or 0)),
    cache_creation_tokens=max(0, int(getattr(usage, "cache_creation_input_tokens", 0) or 0)),
    cache_read_tokens=max(0, int(getattr(usage, "cache_read_input_tokens", 0) or 0)),
)
```

(signature extension belongs in `usage.py` alongside the attribution fields in my earlier comment.)

arksim/llms/chat/providers/openai.py
```diff
@@ -49,6 +50,16 @@ def _prepare_params(
 
         return params
 
+    def _track_response_usage(self, response: object) -> None:
+        usage = getattr(response, "usage", None)
+        if usage:
+            track_usage(
+                self.model,
+                "openai",
+                usage.input_tokens,
```
itsarbit: **Same `None`-attribute risk + missing cached/reasoning token splits.** `usage.input_tokens` is accessed raw. On tool-call-only or refusal responses OpenAI can return missing/None fields; also the Chat Completions → Responses API transition renamed several fields, so attribute access is fragile across SDK versions.

Beyond the None guard, the OpenAI Responses API exposes:
- `response.usage.input_tokens_details.cached_tokens` — prompt cache hits
- `response.usage.output_tokens_details.reasoning_tokens` — o1/o3/gpt-5-reasoning

Both are billed at different rates from standard input/output and are currently invisible. Practitioners running on reasoning models will see reported output tokens under-count by 2-10x.

Same fix shape as the anthropic comment.

arksim/llms/chat/providers/anthropic.py
```diff
@@ -93,9 +104,11 @@ def call(
         # For structured output, return the parsed output
         if schema:
             response = self.client.messages.parse(**params)
+            self._track_response_usage(response)
```
itsarbit, line 118: **Streaming responses silently record zero.** `_track_response_usage(response)` is called on what could be a streaming iterator (when `stream=True` is forwarded via `**kwargs`). `getattr(iterator, "usage", None)` returns `None`, the call returns with no record, and every streamed token is invisible to the tracker. This is a data-integrity bug against the stated purpose of the feature.

Two options:
1. Short-circuit: inspect for `stream` in kwargs and raise / log a warning that streaming is not supported for tracking yet.
2. Consume the stream's final usage event: Anthropic emits `message_delta` with `usage` at end-of-stream; same for OpenAI and Google. The tracker would need a streaming-aware variant that wraps the iterator.

Same pattern at lines 111, 135, 139, and mirrored in the OpenAI / Google providers.

arksim/__init__.py
```diff
@@ -29,11 +29,16 @@
     "Simulator",
     "SimulationInput",
     "SimulationParams",
+    "run_simulation",
+    "UsageTracker",
```
itsarbit: **Public-surface shape is inconsistent.** `UsageTracker` and `UsageRecord` are added to top-level `__all__`, but `TokenUsage` — the Pydantic shape actually attached to `Simulation.usage` / `Evaluation.usage` and visible in serialized outputs — is **not** exported. Users calling `from arksim import ...` cannot annotate the field they will see.

Pre-1.0 breaking-change risk: `UsageTracker` is a `@dataclass` with a `threading.Lock` private field and a public `records: list` that callers should not poke at. Once it's in `__all__`, swapping the lock primitive or renaming `records` is SemVer-major.

Suggested: export `TokenUsage` (the serialized shape, stable) and keep `UsageTracker` / `UsageRecord` under `arksim.llms.chat.base.usage` for advanced users who need them.

Also: `run_simulation` was added to `__all__` on line 32 — unrelated to token tracking, scope creep on this PR.

arksim/utils/html_report/report_template.html
```diff
@@ -1622,6 +1622,15 @@ <h4>Unique Errors</h4>
             </div>
         </div>
 
+        <!-- Token Usage Section -->
+        <div id="tokenUsageSection" style="display:none; margin-top:40px;">
+            <div class="section-title">
+                <span class="material-icons" style="font-size:20px;">data_usage</span>
+                Token Usage
```
itsarbit, line 1706: **Token Usage report section is thin.** Four issues:

1. No units disclosure. A first-time viewer doesn't know if these numbers are tokens (this PR) or USD (common assumption). One subtitle line under the heading ("Totals across all LLM calls in this run; excludes retries that failed before usage metadata was returned") carries the whole section.
2. No per-model breakdown rendered, even though `by_model: dict[str, dict[str, int]]` is collected in `usage.py:summary()` and copied to the report data at `generate_html_report.py:317`. The data is in the JSON but invisible to the viewer.
3. No $ estimate. Token counts without cost are half the feature.
4. Zero tokens displayed is ambiguous: "provider returned no usage metadata" (common on streamed / cache / error paths) looks identical to "no LLM calls made in that stage".

Minimum: add a per-model table underneath the aggregate, one row per `provider/model` key, with columns `input / cached / output / reasoning / total`. Cost estimation can be a fast-follow.

tests/unit/test_usage_tracker.py
```diff
@@ -0,0 +1,231 @@
...
+        tracker = usage.UsageTracker()
+        token = usage.set_current_tracker(tracker)
+        try:
+            usage.track_usage("claude-3-5-sonnet", "anthropic", 50, 20)
+            assert tracker.total_input_tokens == 50
+            assert tracker.total_output_tokens == 20
+        finally:
+            usage.reset_current_tracker(token)
+
+    def test_nested_tracker_not_supported(self, usage: types.ModuleType) -> None:
```
itsarbit: **Test name misleads about behavior.** `test_nested_tracker_not_supported` suggests nesting raises or is forbidden; the body asserts nesting **is** supported (inner shadows outer, outer gets nothing). If the intent is "nested scopes silently replace", rename to `test_nested_tracker_inner_shadows_outer`. If the intent is "nested scopes should be rejected", have `set_current_tracker` raise when a tracker is already active. Either way the current name-vs-behavior split will cost a future reader a minute of double-take.

arksim/llms/chat/providers/anthropic.py
xinyang-articulate: Since we are a toolkit which produces a final report, streaming does not actually help/needed during the process, which is also the reason that we do not have streaming implemented. Will not do further processing with this comment.

arksim/llms/chat/base/usage.py
xinyang-articulate: Not sure about when SDK will return -1 as token count, but applied both fixes

arksim/utils/html_report/report_template.html
xinyang-articulate: Thanks for the advice. I will make add more details to the table about token tracking.

itsarbit (changes requested):
Thanks for this. The core design (ContextVar scope + label stack + explicit `copy_context().run` propagation) is sound, and the backward-compat plumbing on `Simulation.usage` / `Evaluation.usage` is clean. Findings below grouped by severity. Two BLOCKERs and a handful of HIGHs to address before merge. MEDIUMs and LOWs are posted as inline comments on the relevant lines.

I ran the review across six lanes (architect / first-timer / adversary / domain / future-maintainer / Python concurrency) and consolidated.

## BLOCKERs

**1. Azure OpenAI provider not instrumented.**
`AzureOpenAILLM` (`arksim/llms/chat/providers/azure_openai.py`) is a fully supported provider via the `arksim[azure]` extra, but no `_track_response_usage` was added there. Any user on Azure runs the entire simulation and evaluation with zero token tracking, indistinguishable from a tracker-not-active failure. Mirror `OpenAILLM._track_response_usage` (the Responses API surface is identical) and add a unit test. Confirmed by `git diff main..origin/feat/token-tracking -- arksim/llms/chat/providers/azure_openai.py` returning empty.

**2. PR description claims top-level re-exports that do not exist.**
The PR body says `arksim.UsageTracker` and `arksim.UsageRecord` are exported. `arksim/__init__.py:32-38` only lists `TokenUsage` in `_LAZY_IMPORTS`. A user copy-pasting `from arksim import UsageTracker` hits `AttributeError`. Either fix the description (TokenUsage-only is the safer choice given public-API commitment costs) or actually export `UsageTracker`, `UsageRecord`, `usage_scope`, `usage_label`, `track_usage`. Pick one and align both.

**3. No README, CHANGELOG, or `docs/` entry.**
`git diff main..origin/feat/token-tracking -- '*.md'` is empty. The PR self-checks "No docs needed" because the feature is automatic, but "automatic capture" is half the value. The other half is a user reading the data: the JSON schema for `simulation.usage` / `evaluation.usage`, and the canonical CI-budget recipe. Add:
- `CHANGELOG.md` entry under `## [Unreleased] → Added`.
- A "Token Usage" section in `README.md` or `docs/main/` listing the output schema (`total_input_tokens`, `total_output_tokens`, `total_cached_tokens`, `total_reasoning_tokens`, `by_model`, `breakdowns.by_conversation`) and a 10-line CI-budget snippet using `jq '.usage.total_input_tokens'`.

## HIGHs

**4. Anthropic input-token math conflates billable categories.**
`arksim/llms/chat/providers/anthropic.py:78-89`. The code computes `total_input = base_input + cache_read + cache_create` and surfaces `cached_tokens=cache_read` only. Anthropic bills these at 1x, 0.1x, and 1.25x rates respectively. Two consequences:
- `cache_create` is folded into `input_tokens` but not surfaced as a separate field, so a downstream cost calc using `(input - cached) * input_rate + cached * cache_read_rate` will both over-bill cache_create (rated at 1x instead of 1.25x) and incorrectly include cache_read in the residual.
- OpenAI's `input_tokens_details.cached_tokens` is already part of `usage.input_tokens` (semantics differ from Anthropic). Reporters comparing providers double-count.

Fix options: (a) surface three explicit fields (`input_tokens` uncached, `cached_input_tokens` for cache_read, `cache_creation_input_tokens` for cache_create), or (b) normalize all providers to the same convention and document it loudly. (a) is the cleaner path because it preserves the information needed for cost.

**5. No cost surface, and the data model makes adding it later painful.**
Competitors (LangSmith, Braintrust, Maxim) lead with `$ cost`, not tokens. `TokenUsage` persists only top-level aggregates; raw `records` are dropped at serialization. A future `cost_usd` either has to be recomputed from records (gone after persistence) or forces a v1.2 schema bump. The `breakdowns` field already exists on `TokenUsage` but is not wired through the serialized output. Persist either `records: list[UsageRecord]` or pre-computed `breakdowns.by_model` so the cost layer can attach without breaking persistence.

**6. `copy_context()` called per `executor.submit` inside a comprehension.**
`arksim/evaluator/evaluator.py:288, 331` and `arksim/evaluator/evaluate.py:172-181`. Each worker gets an independent context snapshot. Today the comprehension does not mutate `_current_labels` between iterations, so it is safe by accident. Adding a per-turn `usage_label(turn_id=...)` inside the comprehension would silently desync. Lift `ctx = contextvars.copy_context()` once outside, reuse for all submits. Better: extract `arksim/utils/concurrency.py:submit_with_context(pool, fn, *args)` so future telemetry features (latency, retries, cache hits) reuse the same primitive.

**7. `UsageRecord` is a mutable dataclass exposed via `tracker.records`.**
`arksim/llms/chat/base/usage.py:40-51`. Any caller iterating `tracker.records` can mutate `input_tokens` after `record()` appends, bypassing the tracker's lock. Apply `@dataclass(frozen=True)`. No field needs post-construction mutation.

**8. No test exercises `asyncio.create_task` propagation.**
`tests/unit/test_usage_tracker.py::TestConcurrentTracking` covers only `ThreadPoolExecutor`. The simulator's `asyncio.create_task` path is correct today only because `track_usage()` mutates a shared aggregator object reference rather than rebinding the ContextVar. One refactor (e.g. switching to `asyncio.to_thread`, which does not propagate context to the thread by default) silently loses tokens. Add `test_asyncio_create_task_propagates_to_parent_tracker` and a documented anti-pattern test for `asyncio.to_thread`.

**9. No provider-level unit tests for `_track_response_usage`.**
None of the three helpers has a test for missing `usage` field, `usage.input_tokens=None`, or per-provider edge shapes. Each provider's null-handling differs (OpenAI uses chained `getattr`, Anthropic does not, Google reads from `usage_metadata`). Add one test per provider with a recorded SDK response fixture, including the null and missing cases. With `clean_usage_value` swallowing `TypeError` / `ValueError` to 0 (see inline note on `usage.py:17`), absent tests means a provider response-shape change reports 0 silently.

**10. Parallel pipeline to OTel GenAI semconv.**
arksim already runs an OTel / OpenInference pipeline (`span_converter.py` reads `gen_ai.tool.*`). This PR captures the same data in a `ContextVar` next to provider SDK calls. Not wrong, but two tracking pipelines is a smell. At minimum confirm `UsageRecord.provider` aligns with `gen_ai.system` and `input_tokens` / `output_tokens` align with the semconv attribute names (they do, keep it that way). In a follow-up, consider sourcing `UsageTracker` from spans so OpenInference instrumentation auto-feeds it.

## Closing notes

- HTML report XSS verified safe (provider / model names and breakdown values pass through `escapeHtml`).
- Backward compat verified: `Simulation.usage: TokenUsage | None = None` defaults correctly; old JSON parses.
- Commit messages on the branch (`add token tracking`, `fix bug`, `fix tests`) make `git blame` non-recoverable in 6 months. Squash with the PR-title commit message body capturing the design (ContextVar scope, label stack, copy_context propagation, schema additions).
- Remaining MEDIUMs and LOWs are inline on the relevant lines (clean_usage_value silent zeros, schema versioning, test isolation, streaming gotcha, usage_scope nesting, pre-existing messages.parse concern, pattern-propagation docstring, public-name choice, threading.Lock + replace).

Happy to pair on any of the BLOCKERs or HIGHs. The Anthropic input-token semantics (#4) is the one most likely to silently mis-bill production once a cost layer attaches, so worth nailing down before merge even if cost itself ships in a follow-up.

arksim/llms/chat/base/usage.py
```diff
@@ -0,0 +1,290 @@
...
+from contextvars import ContextVar, Token
+from dataclasses import dataclass, field
+
+from pydantic import BaseModel
+from pydantic import Field as PydanticField
+
+logger = logging.getLogger(__name__)
+
+
+def clean_usage_value(value: object) -> int:
```
itsarbit, line 18: **MEDIUM:** swallows `TypeError` and `ValueError` to `0` silently. If a provider response shape changes, every call reports 0 with no log line and oncall sees 'tokens went to zero' with no signal. Add `logger.warning` once per unexpected type per process so the regression is detectable from logs.

arksim/simulation_engine/entities.py
```diff
@@ -186,3 +187,4 @@ class Simulation(BaseModel):
         )
     )
     conversations: list[Conversation]
+    usage: TokenUsage | None = None
```
itsarbit, line 190: **MEDIUM:** schema versioning. `Simulation.usage` was added without a `schema_version` field. The next required field has no migration path. Consider adding `schema_version: int = 1` now (cheap, backward-compatible) or a `# TODO(schema-versioning)` comment with a follow-up issue link. Same applies to `Evaluation.usage` in `arksim/evaluator/entities.py:252`.

arksim/evaluator/entities.py
```diff
@@ -248,4 +249,5 @@ class Evaluation(BaseModel):
     simulation_id: str
     conversations: list[ConversationEvaluation]
     unique_errors: list[UniqueError]
+    usage: TokenUsage | None = None
```
itsarbit, line 273: **MEDIUM:** see the schema-versioning note on `Simulation.usage` at `arksim/simulation_engine/entities.py:190`. Same recommendation here.

tests/unit/test_usage_tracker.py
```diff
@@ -0,0 +1,447 @@
...
+        finally:
+            usage.reset_current_tracker(t2)
+            usage.reset_current_tracker(t1)
+
+
+# ---------------------------------------------------------------------------
+# Threads without copy_context() do NOT inherit the parent tracker
+# ---------------------------------------------------------------------------
+class TestThreadIsolation:
+    def test_raw_thread_does_not_inherit_tracker(self, usage: types.ModuleType) -> None:
```
itsarbit, line 276: **MEDIUM:** test isolation. This class mutates the module-scoped `_current_tracker` ContextVar outside `try/finally`. If an assertion in `test_raw_thread_does_not_inherit_tracker` fails before the reset, state leaks into subsequent tests in the same module. Add an `autouse` fixture in `tests/conftest.py` that snapshots and resets `_current_tracker` and `_current_labels` per test, or wrap each setter in `try/finally`.

arksim/llms/chat/base/usage.py
```diff
@@ -0,0 +1,290 @@
...
+    provider: str
+    input_tokens: int
+    output_tokens: int
+    cached_tokens: int = 0
+    reasoning_tokens: int = 0
+    labels: dict[str, str] = field(default_factory=dict)
+
+
+@dataclass
+class UsageTracker:
```
itsarbit: **MEDIUM:** streaming silent zero. None of the three providers use streaming today, but `**kwargs` is accepted on `call` / `call_async`. Any caller passing `stream=True` produces zero usage with no error. Either reject streaming kwargs explicitly in `_prepare_params`, or add a line in this class's docstring: 'Streaming responses must enable provider-side usage emission (e.g. `stream_options={"include_usage": True}` for OpenAI, or read the `message_stop` event for Anthropic). Otherwise totals are zero.'

arksim/llms/chat/base/usage.py
```diff
@@ -0,0 +1,290 @@
...
+    return _current_tracker.set(tracker)
+
+
+def reset_current_tracker(token: Token[UsageTracker | None]) -> None:
+    """Reset the tracker context variable using the token from set_current_tracker."""
+    _current_tracker.reset(token)
+
+
+@contextmanager
+def usage_scope() -> Iterator[UsageTracker]:
```
itsarbit: **MEDIUM:** nesting behavior is unspecified in the docstring. An inner `usage_scope()` shadows the outer for the inner's duration; the outer never sees inner records. Add a one-line note: 'Scopes do not aggregate upward. Do not nest `usage_scope` around `run_simulation` / `run_evaluation`, which manage their own scope and expose totals on `Simulation.usage` / `Evaluation.usage`.'

arksim/llms/chat/providers/anthropic.py
```diff
@@ -93,9 +112,11 @@ def call(
...
```
itsarbit, line 118: **MEDIUM (pre-existing, not introduced by this PR):** the Anthropic Python SDK as of late 2025 does not expose `messages.parse()`. Either there is an undocumented wrapper, this code path is unused, or the structured-output branch has been silently broken. Worth filing a separate issue. Flagging here because this PR makes the gap more visible by wiring `_track_response_usage` immediately after the call.

arksim/llms/chat/base/usage.py
```diff
@@ -0,0 +1,290 @@
...
+    model: str
+    provider: str
+    input_tokens: int
+    output_tokens: int
+    cached_tokens: int = 0
+    reasoning_tokens: int = 0
+    labels: dict[str, str] = field(default_factory=dict)
+
+
+@dataclass
```
itsarbit: **MEDIUM:** this PR establishes 'instrument all providers via a ContextVar' as a pattern, but the convention is not named anywhere. Future telemetry features (latency, retries, cache hits) will need the same propagation primitives, and without guidance someone will sprinkle `usage_label()` into utility helpers and create implicit coupling. Suggest a 6-line module docstring at the top of `usage.py`:

- only provider adapters call `track_usage`
- only top-level entry points (`run_simulation`, `run_evaluation`, manual user code) open `usage_scope`
- `usage_label` is for orchestration boundaries, not leaf functions
- thread spawns must use `contextvars.copy_context().run`

arksim/__init__.py
```diff
@@ -29,11 +29,13 @@
     "Simulator",
     "SimulationInput",
     "SimulationParams",
+    "TokenUsage",
     "ToolCall",
     "ToolCallSource",
 ]
 
 _LAZY_IMPORTS: dict[str, tuple[str, str]] = {
+    "TokenUsage": (".llms.chat.base.usage", "TokenUsage"),
```
itsarbit, line 38: **LOW:** `UsageTracker` as a public-API name reads as plumbing to first-time users. LangSmith / Braintrust / Maxim expose 'usage' / 'metrics' / 'cost' as the user-facing entry point. If you do export the tracker (see BLOCKER #2 in the review body), consider promoting `usage_scope()` as the first-class entrypoint and demoting `UsageTracker` to internals. The contextmanager is what 95% of callers actually want.

arksim/llms/chat/base/usage.py
```diff
@@ -0,0 +1,290 @@
...
+        with usage_scope() as tracker:
+            with usage_label(component="conversation", conversation_id="c1"):
+                result = run_simulation(...)
+
+        print(tracker.total_input_tokens, tracker.total_output_tokens)
+        print(tracker.summary_by("conversation_id"))
+    """
+
+    records: list[UsageRecord] = field(default_factory=list)
+    _lock: threading.Lock = field(
```
itsarbit, line 75: **LOW:** the `_lock` field plus a dataclass means `dataclasses.replace(tracker)` produces a clone with an independent lock that is not thread-safe against the original. Currently nothing in the codebase calls `replace`, but plausible in tests/snapshots. Either document non-copyable, or implement `__deepcopy__` and `__copy__`. `field(compare=False, repr=False)` is already correct, just missing the copy semantics.

PR #140, opened 2026-04-08
feat(a2a): add tool call capture via DataPart

itsarbit:
## Summary

- Extract tool calls from `DataPart` in A2A protocol responses so the evaluator can score tool usage with `tool_call_behavior_failure`
- Add `source` field to `ToolCall` for capture provenance tracking (`a2a_protocol`)
- Add `ToolCallSource` constant class to avoid magic strings
- Harden `_extract_response` with `isinstance` guards for malformed remote data
- Document the expected DataPart JSON schema in `tool-call-capture.mdx`
- Add complete working example with OpenAI Agents SDK + A2A server

## How it works

The remote A2A agent embeds tool call data as a `DataPart` alongside its `TextPart` text answer:

```json
{"tool_calls": [{"id": "call_1", "name": "get_weather", "arguments": {"city": "Tokyo"}, "result": "72F, sunny"}]}
```

arksim's A2A client extracts these automatically into `AgentResponse.tool_calls`, which the evaluator uses for trajectory matching.

## Test plan

- [x] 15 unit tests covering happy path, edge cases, and adversarial inputs (malformed DataPart data)
- [x] Full unit suite: 751 passed, 0 failures
- [x] End-to-end manual test: started example A2A server, ran `arksim simulate-evaluate`, verified tool calls appear in simulation output with `source=a2a_protocol` and flow through to evaluator scoring
- [x] Pre-commit hooks pass (ruff, bandit, license headers)


examples/tool-call-capture/a2a/agent_server/agent.py
```diff
@@ -0,0 +1,126 @@
...
+
+_sdk_agent = SDKAgent(
+    name="WeatherAgent",
+    instructions=_SYSTEM_INSTRUCTIONS,
+    tools=[get_weather],
+    model="gpt-4.1-mini",
+)
+
+
+def _extract_tool_calls(result: RunResult) -> list[dict[str, Any]]:
```
nrajanee: is this something every customer will have to implement on their own? can we make it a part of our code instead?

shehan360 (changes requested):
Checked the A2A spec on how tool call results should be conveyed. Section 3.7 (Messages and Artifacts) is explicit:

> "Messages SHOULD NOT be used to deliver task outputs. Results SHOULD BE returned using Artifacts associated with a Task."

The spec draws a hard line: Messages are for communication (task initiation, clarification, status). Artifacts are for output data. Tool call results are task output, so they belong in `Task.artifacts`, not embedded in a Message.


Something to note:

A2A also deliberately defines no tool-call semantics of its own ("A2A ❤️ MCP" doc: A2A is for agent collaboration, MCP is for tool invocation). So any schema here is a convention we're layering on top. The spec's recommended way to do that is an **AgentExtension** with a URI, declared in the AgentCard, with payload carried in the `metadata` map of a core data structure. That gives us a namespaced, negotiable, versionable convention instead of a bare reserved artifact name.

Refs: [A2A spec §3.7](https://a2a-protocol.org/latest/specification/), [Extensions guide](https://a2a-protocol.org/latest/topics/extensions/), [A2A and MCP](https://a2a-protocol.org/latest/topics/a2a-and-mcp/).

itsarbit:
@shehan360 great catch on the spec violation — refactored per your feedback.

Switched from `DataPart`-in-`Message` to a versioned `AgentExtension` (`https://arksim.arklex.ai/a2a/tool-call-capture/v1`) declared in `AgentCard.capabilities.extensions`, with tool calls carried in `Task.artifacts[*].metadata` keyed by `{URI}/tool_calls` per the konami-code convention from the extensions doc.

Client negotiates via the SDK's `extensions=` parameter and reads from `Task.artifacts` per §3.7. Messages carry no tool outputs. Handles streaming `TaskArtifactUpdateEvent` + final `Task` snapshot, plus `TaskStatusUpdateEvent` as no-op.

Verified end-to-end: AgentCard declares the extension, server calls `add_activated_extension` so `x-a2a-extensions` echoes back in response headers, tool calls flow through with `source=a2a_protocol`.

examples/tool-call-capture/a2a/agent_server/agent.py
itsarbit: Good call — extracted into `arksim.tracing.openai.extract_tool_calls` as a reusable library helper. Both the A2A example and `examples/customer-service/custom_agent.py` now import it instead of duplicating the logic.

nrajanee:
lgtm!

nrajanee (dismissed):

shehan360 (changes requested):
Three small requested changes, plus a security question we should discuss as we need the guidance story right before many users start writing servers against this extension.

## Requested changes

**1. Fold the example into `examples/customer-service/`**

Our `examples/` layout is domain-based: `bank-insurance`, `customer-service`, `e-commerce`, `openclaw`. `examples/tool-call-capture/a2a/` breaks that by naming a folder after a feature, and it leaves the only A2A example in the repo as a throwaway weather toy rather than a realistic agent.

Suggested restructure:

```
examples/customer-service/
├── custom_agent.py          # existing Python connector
├── traced_agent.py          # existing Python + trace receiver
├── a2a_server/              # NEW
│   ├── agent.py             # reuses customer-service tools
│   ├── server.py            # AgentCard + extension + TaskUpdater wiring
│   └── README.md
├── config_a2a.yaml          # NEW
└── scenarios.json           # shared
```

Factor the SQLite-backed tools in `custom_agent.py` into a `tools.py` that both the Python connector and the new A2A server import. Readers then get one domain, one set of scenarios, and three transport flavors (Python, Python+trace, A2A) of the same agent. The A2A-specific wiring stays small and readable because tool definitions live elsewhere. Doc link in `tool-call-capture.mdx` updates to point at the new path, and `examples/tool-call-capture/` goes away.

**2. Clarify or drop `name="response"` in the example artifact**

`Artifact.name` is optional and purely a human-readable label for multi-artifact tasks. arksim doesn't consume it; the client filters on `artifact.extensions` and reads `artifact.metadata`, not `artifact.name`. Either drop `name="response"` from the example, or add an inline comment so readers don't think it's contractual.

**3. Rephrase the Message/DataPart warning**

The current Warning reads as a reaction to our earlier review thread. DataPart-in-Message is a natural first guess for anyone reading A2A docs cold (it's a valid Part type, and the spec doesn't shout "outputs don't go here" from the top of the page), so the Warning earns its spot on pedagogical grounds. The tone could be neutral and spec-anchored:

> Arksim's A2A client reads `Task.artifacts` only (per A2A spec §3.7). Tool call data embedded in `Message.parts`, including `DataPart`, will not be captured.

## Security: open question for discussion

Raising this for thinking. The PR ships working code, but the guidance we publish alongside it shapes how users deploy servers against this extension, and I want to make sure we get the story right before it calcifies.

**The issue.** In the example server, tool call metadata is added to the artifact unconditionally:

```python
if A2AToolCaptureExtension.URI in context.requested_extensions:
    context.add_activated_extension(A2AToolCaptureExtension.URI)

# ...

if tool_calls:
    metadata[A2AToolCaptureExtension.METADATA_KEY] = tool_calls
    extensions.append(A2AToolCaptureExtension.URI)
```

The negotiation check only controls the response header echo. Any A2A client hitting this endpoint (arksim or anyone else) receives the full tool call trace in artifact metadata. That matters because tool calls tend to expose internals the LLM text response deliberately avoids: raw arguments (PII, customer IDs, verification codes, tokens), tool names (recon surface), and verbatim tool results. A server author who drops this example onto a reachable endpoint is exposing strictly more than they would without the extension.

Worth noting: even if we gate emission on `requested_extensions`, that is still just a request-header opt-in. Any HTTP client can send `A2A-Extensions: <URI>`. Negotiation is a signaling mechanism, not access control.

So the question is what we ask users to do. A few deployment options, ordered by isolation strength:

**Option A: Dedicated evaluation deployment.** Two instances of the agent: a prod-facing one that does not declare the extension at all, and an eval-only one behind private networking that does. Zero risk of leaking to prod traffic, at the cost of running a second process. Best fit for regulated domains (HIPAA, PCI) or when prod and eval are operated by different teams.

**Option B: Single deployment with transport auth plus negotiation gate.** One endpoint, behind bearer or mTLS auth declared in `AgentCard.security_schemes`. arksim carries the credential via `A2AConfig.headers` (the plumbing is already there in `a2a.py`). The server gates tool call emission on both (a) valid auth at the transport layer and (b) the `A2A-Extensions` request header declaring the URI. Auth keeps strangers out; the negotiation gate provides least-privilege within the authenticated group, so internal services with tokens that don't need tool calls don't receive them. Likely the common case for most teams.

**Option C: Env-flag gate.** Server only emits the extension when `ARKSIM_TOOL_CAPTURE_ENABLED=true` at startup. Simplest but weakest. Useful for CI/staging where the whole deployment is eval-facing; doesn't help a mixed-use endpoint, and a flipped flag in prod silently turns on exposure.

**My suggestion for where to land.** Some combination of:

1. Update the example server to gate emission on `requested_extensions` in addition to the header echo. Three-line change, aligns with A2A's negotiation semantics, means the example models the right default posture even when users don't read the docs carefully.

2. Add a Security section to `tool-call-capture.mdx` calling out the threat model explicitly (tool call metadata is more sensitive than the response text), describing the three deployment options above, and showing the `AgentCard.security_schemes` + `A2AConfig.headers` pattern for Option B since that is what most users will land on.

3. Leave deeper options (mTLS config, dedicated eval deployment topology, secret rotation) as pointers rather than fully worked examples.

Would love your read: does this framing feel right, or are there deployment patterns we should add? Happy to pair on the doc writeup once we align on the shape.

itsarbit:
@shehan360 addressed all three items plus the security discussion:

**1. Folded into customer-service.** Extracted shared tools into `tools.py` (6 tools, DB, instructions). New `a2a_server/` sits alongside `custom_agent.py` and `traced_agent.py` — same domain, same scenarios, three transports. Old `examples/tool-call-capture/` deleted.

**2. Dropped `name="response"`.** Removed from docs snippet and server.

**3. Rephrased warning.** Now uses your suggested wording: "arksim's A2A client reads `Task.artifacts` only (per A2A spec §3.7)..."

**4. Emission gated on `requested_extensions`.** Server checks `A2AToolCaptureExtension.URI in context.requested_extensions` before emitting tool call metadata. Calls `add_activated_extension` so the header echoes back. Clients that don't request the extension get clean artifacts with no tool trace data.

**5. Security section added** to `tool-call-capture.mdx`. Covers the threat model (tool call metadata exposes more than text), three deployment options ordered by isolation (dedicated eval, auth+negotiation, env flag), and the `A2AConfig.headers` pattern for Option B.

On the `securitySchemes` server-side snippet — I left that as a pointer to the A2A spec rather than a worked example, since the auth setup is deployment-specific. Happy to add it if you think the docs need both halves.

Verified end-to-end: 28 tool calls captured across all 6 customer-service tools via the A2A path.

shehan360 (dismissed):
A couple of follow-ups after rereading the merged changes end-to-end:

### 1. The `Server requirements` snippet teaches the ungated pattern

The example server (`examples/customer-service/a2a_server/server.py`) correctly gates tool call emission on the client's `A2A-Extensions` request header:

```python
emit_tool_calls = A2AToolCaptureExtension.URI in context.requested_extensions
if emit_tool_calls:
    context.add_activated_extension(A2AToolCaptureExtension.URI)
# ... later ...
if emit_tool_calls and tool_calls:
    metadata[A2AToolCaptureExtension.METADATA_KEY] = tool_calls
    extensions.append(A2AToolCaptureExtension.URI)
```

But the canonical snippet in `docs/main/tool-call-capture.mdx` under "Server requirements" (the one a developer will copy when implementing the extension) emits unconditionally:

```python
await updater.add_artifact(
    parts=[Part(root=TextPart(text=answer))],
    metadata={A2AToolCaptureExtension.METADATA_KEY: [...]},
    extensions=[A2AToolCaptureExtension.URI],
)
```

A developer reading linearly sees the emission snippet first and the Security section later. If they stop after the copy-paste, their server leaks tool call metadata to every A2A client that can reach the endpoint. The Security section mentions the example gates on `requested_extensions`, but the code they'd paste into their executor doesn't show that gate.

Could we inline the `requested_extensions` check into the canonical snippet, with a one-line comment pointing at `#security-considerations` for the rationale? That way the copy-paste path teaches the safe pattern by default, and the Security section expands on why and on the deployment options around it.

### 2. The extension URI should resolve to something human-readable

`https://arksim.arklex.ai/a2a/tool-call-capture/v1` is used as a string identifier per the A2A extensions convention, but it's also a real URL that people will paste into a browser the moment they see it in their AgentCard, in the `A2A-Extensions` response header, or in debug logs. Right now it 404s (or whatever the current `arksim.arklex.ai` default is).

Two reasonable options:

- **Redirect** (`302`) to the Mintlify `tool-call-capture` page's `#a2a-protocol-capture` anchor. Simplest and keeps a single source of truth.
- **Serve a small static page** at that path with the schema, version history, and a link to the full docs. Better for archival/versioning once we ship `/v2`, since `/v1` can keep describing v1 after the docs move on.

Either is fine for v1; I'd lean toward the redirect now and promote to a static page when we bump the version. The A2A extensions spec ([link](https://a2a-protocol.org/latest/topics/extensions/)) explicitly recommends hosting docs at the extension URI for exactly this discoverability reason, and using `w3id.org` for permanence if we want the URI to outlive any future domain changes — worth considering but not blocking.

itsarbit:
@shehan360 addressed both follow-ups:

**1. Snippet now teaches the gated pattern.** The "Server requirements" code block in the docs now includes the `requested_extensions` check and `add_activated_extension` call inline, with a comment pointing to `#security-considerations`. Copy-paste gives users the safe default.

**2. Extension URI resolution.** Agreed on the 302 redirect to the Mintlify page for now. That's a DNS/hosting change outside this PR. Will track separately.

Also rebased onto latest main (includes v0.3.4 release). 782 tests pass.

shehan360 (dismissed):

shehan360 (approved):

PR #162, opened 2026-04-23
feat(evaluator): add scope (turn vs. conversation) to custom metrics

KathyLiu20:
## Summary

- Adds a `scope` parameter (`"turn"` | `"conversation"`) to `QuantitativeMetric` and `QualitativeMetric` so custom metrics can declare whether they run per agent response or once per full conversation. Default is `"turn"` for backward compatibility.
- Renames `evaluate_goal_completion` → `evaluate_conversation` and extends it to run conversation-level custom metrics alongside goal completion scoring.
- Routes loaded custom metrics by scope in `run_evaluation`, passing turn-level metrics to `evaluate_turn` and conversation-level metrics to `evaluate_conversation`.
- Adds `scope` field to `QuantResult` / `QualResult` so thresholds and the HTML report can identify metric origin without re-querying.
- Updates `check_numeric_thresholds` to use the per-conversation score directly for conversation-scope metrics (instead of averaging turn scores).
- Updates `check_qualitative_failure_labels` to also check `convo.qual_scores`.
- Redesigns the HTML report transcript modal scores section as an expandable list with inline reasons; qual drill-down surfaces conversation-level matches as clickable conversation chips.
- Updates docs (`evaluate-conversation.mdx`) with scope semantics and examples.
- Updates the bank-insurance example to annotate all 7 custom metrics with explicit `scope=`.

## Test plan

- [x] `pytest tests/unit/` passes (new tests cover `evaluate_conversation` custom metric execution, `additional_input` forwarding, and scope-based metric loading)
- [x] `arksim simulate-evaluate examples/bank-insurance/config.yaml` runs end-to-end without errors
- [x] HTML report shows conversation-level metrics in the transcript modal and qual drill-down
- [x] Numeric and qualitative threshold gates apply correctly for both turn- and conversation-scope metrics
- [x] Existing turn-level custom metrics continue to work unchanged (backward compat)

## Updated report screenshots
conversation modal shows all conversation level metric scores
Screenshot 2026-04-23 at 4 19 18 PM

expand (conversation-level) metric to see the linked conversations
Screenshot 2026-04-23 at 4 19 42 PM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Deprecated
+
+* **evaluator:** `EvaluationParams.custom_qualitative_metrics` is deprecated; pass qualitative metrics in `custom_metrics` alongside quantitative ones. The field still works but emits a `DeprecationWarning` and will be removed in a future major release.
+
 ### Changed
 
 * **a2a:** migrate from a2a-sdk 0.3.x (Pydantic) to 1.0.0 (protobuf) ([#152](https://github.com/arklexai/arksim/pull/152))
--- a/arksim/evaluator/__init__.py
+++ b/arksim/evaluator/__init__.py
@@ -9,6 +9,7 @@ from arksim.scenario.entities import AssertionType, ToolCallsAssertion
 
 from .base_metric import (
     ChatMessage,
+    MetricScope,
     QualitativeMetric,
     QualResult,
     QuantitativeMetric,
@@ -34,6 +35,7 @@ __all__ = [
     "ChatMessage",
     "Evaluation",
     "EvaluationInput",
+    "MetricScope",
     "QuantitativeMetric",
     "QualitativeMetric",
     "QualResult",
--- a/arksim/evaluator/base_metric.py
+++ b/arksim/evaluator/base_metric.py
@@ -31,6 +31,7 @@ class QuantResult(BaseModel):
     value: float
     reason: str | None = None
     metadata: dict[str, Any] | None = None
+    scope: str = "turn"
 
 
 class QualResult(BaseModel):
@@ -39,6 +40,7 @@ class QualResult(BaseModel):
     name: str
     value: str
     reason: str | None = None
+    scope: str = "turn"
 
 
 class ScoreInput(BaseModel):
@@ -100,6 +102,9 @@ class QuantitativeMetric(_LLMMixin, abc.ABC):
             metrics). The range is used to normalise the value for the aggregated
             turn score and to derive a per-metric failure threshold
             (60 % of *max*).
+        scope: When the metric runs: ``"turn"`` (default) evaluates every
+            individual turn; ``"conversation"`` evaluates once per conversation
+            alongside goal completion.
         llm: Optional LLM instance injected by the evaluator. Custom metrics that
             need an LLM for scoring should accept ``llm=None`` in their
             ``__init__`` and pass it to ``super().__init__(llm=llm)``, then use
@@ -135,12 +140,14 @@ class QuantitativeMetric(_LLMMixin, abc.ABC):
         additional_input: dict[str, Any] | None = None,
         description: str = "",
         llm: BaseLLM | None = None,
+        scope: str = "turn",
     ) -> None:
         self.name = name if name is not None else self.__class__.__name__
         self.score_range = score_range
         self.additional_input = additional_input or {}
         self.description = description
         self._llm = llm
+        self.scope = scope
 
     @abc.abstractmethod
     def score(self, score_input: ScoreInput) -> QuantResult:
@@ -157,6 +164,9 @@ class QualitativeMetric(_LLMMixin, abc.ABC):
         name: The name of the metric. If not provided, uses the class name as default.
         description: Human-readable description of what the metric measures.
         label_colors: Mapping of label values to hex colour strings for the report UI.
+        scope: When the metric runs: ``"turn"`` (default) evaluates every
+            individual turn; ``"conversation"`` evaluates once per conversation
+            alongside goal completion.
         llm: Optional LLM instance injected by the evaluator. Custom metrics that
             need an LLM for evaluation should accept ``llm=None`` in their
             ``__init__`` and pass it to ``super().__init__(llm=llm)``, then use
@@ -191,11 +201,13 @@ class QualitativeMetric(_LLMMixin, abc.ABC):
         description: str = "",
         label_colors: dict[str, str] | None = None,
         llm: BaseLLM | None = None,
+        scope: str = "turn",
     ) -> None:
         self.name = name if name is not None else self.__class__.__name__
         self.description = description
         self.label_colors: dict[str, str] = label_colors or {}
         self._llm = llm
+        self.scope = scope
 
     @abc.abstractmethod
     def evaluate(self, score_input: ScoreInput) -> QualResult:
--- a/arksim/evaluator/base_metric.py
+++ b/arksim/evaluator/base_metric.py
@@ -151,6 +151,12 @@ class QuantitativeMetric(_LLMMixin, abc.ABC):
         self._llm = llm
         self.scope = scope
 
+    def run(self, score_input: ScoreInput) -> QuantResult:
+        """Score *score_input* and stamp ``scope`` on the result."""
+        r = self.score(score_input.model_copy(update=self.additional_input))
+        r.scope = self.scope
+        return r
+
     @abc.abstractmethod
     def score(self, score_input: ScoreInput) -> QuantResult:
         raise NotImplementedError()
@@ -210,6 +216,12 @@ class QualitativeMetric(_LLMMixin, abc.ABC):
         self.label_colors: dict[str, str] = label_colors or {}
         self._llm = llm
         self.scope = scope
+
+    def run(self, score_input: ScoreInput) -> QualResult:
+        """Evaluate *score_input* and stamp ``scope`` on the result."""
+        r = self.evaluate(score_input)
+        r.scope = self.scope
+        return r
 
     @abc.abstractmethod
     def evaluate(self, score_input: ScoreInput) -> QualResult:
--- a/arksim/evaluator/base_metric.py
+++ b/arksim/evaluator/base_metric.py
@@ -210,16 +210,18 @@ class QualitativeMetric(_LLMMixin, abc.ABC):
         label_colors: dict[str, str] | None = None,
         llm: BaseLLM | None = None,
         scope: MetricScope = "turn",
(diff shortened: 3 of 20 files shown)


shehan360 (changes requested):
Nice split. A few suggestions before this lands:

**1. Type `scope` as `Literal`, not `str`**

Every consumer compares `== "turn"` / `== "conversation"` (`evaluator.py:928-935`, `thresholds.py:58`). Today a typo like `scope="conversations"` is silently dropped from both `turn_quant` and `convo_quant` lists — the metric never runs, no warning. `Literal` gives mypy coverage with no runtime cost:

```python
from typing import Literal
MetricScope = Literal["turn", "conversation"]
```

Prefer this over an enum — custom-metric authors keep passing plain strings.

**2. Several naming inconsistencies worth fixing in one pass**

- `EvaluationParams.convo_custom_metrics` / `convo_custom_qualitative_metrics` (`entities.py:165-168`) break the existing `custom_*` prefix pattern. Suggest `custom_convo_metrics` / `custom_convo_qualitative_metrics`.
- `ConversationEvaluation.scores` / `qual_scores` (`entities.py:233-236`) are ambiguously named — a `ConversationEvaluation` already has `turn_scores[*].scores`. A reader can't tell these hold *conversation-scope custom* metrics without the comment. Suggest `convo_scores` / `convo_qual_scores` (or `custom_scores` / `custom_qual_scores`).
- `gc_max_workers`, `gc_inner_workers`, `gc_futures` in `evaluator.py:323-342` — the `gc_` prefix is stale now that this loop does more than goal completion. Rename to `convo_*`. The `# Phase 2: goal_completion for each conversation` comment on line 320 is similarly stale.
- `evaluate_conversation(custom_metrics=..., custom_qualitative_metrics=...)` (`evaluate.py:247-248`) uses the same parameter names as `evaluate_turn` but means conversation-scope metrics. Easy mistake at a call site. If `EvaluationParams` is renamed as above, rename these to match.

**3. Simplification: the `_run`/`_run_qual` wrappers duplicate `evaluate_turn`**

`evaluate.py:287-319` is near-identical to `evaluate.py:97-131` — the only differences are the absence of `current_turn` and the `scope="conversation"` tag. The wrapping exists solely to stamp `scope`, which the metric already knows. Two options:

- Stamp in the base class: `result.scope = self.scope` after `score()`/`evaluate()`, then drop the per-call wrapping.
- Extract a shared `_run_metrics_parallel(metrics, score_input, num_workers)` helper used by both functions.

Minor: `reason=r.reason or ""` coerces `None` → `""` even though `QuantResult.reason` is `str | None`. Drop the coercion.

**4. Test quality**

`TestLoadCustomMetricsScope` in `test_evaluator_class.py:396` wraps assertions in `if quant:` — if loading silently fails, the test passes vacuously. Drop the guards or `assert quant` first.

**5. Could you drop in a screenshot of the new transcript modal / qual drill-down?**

The HTML report changes are hard to review from the template diff alone. A screenshot of the expandable scores list and a convo-level qual drill-down (with both turn-level and convo-level matches, if possible) would help.

shehan360:
A few more from a second pass:

**1. Consider a more composable shape for `EvaluationParams`**

Right now scope knowledge lives in three places: the metric class, the loader/splitter in `run_evaluation`, and the two evaluator entry points. `EvaluationParams` carries four parallel metric lists.

A single `custom_metrics: list[Metric]` (union of quant + qual) on `EvaluationParams`, with each metric's `scope` attribute read at execution time by a unified runner that maps `scope → context`, collapses this. Adding a future scope (`"dataset"`, `"trajectory"`) would then touch 2 files instead of 6.

Not blocking — but worth considering while the API is still being shaped.

**2. `display_evaluation_summary` silently averages turn + convo scores with the same name**

`evaluator.py:707-713`:

```python
for conv in results.conversations:
    for turn in conv.turn_scores:
        for score in turn.scores:
            metric_scores[score.name].append(score.value)  # turn-scope
        ...
    for score in conv.scores:
        metric_scores[score.name].append(score.value)      # convo-scope -- same bucket
```

If a turn-scope and convo-scope metric share a name (user mistake, or two metrics both named `"relevance"`), their values are pooled into the same `metric_scores[name]` list and averaged together — mixing a 0–1 goal-completion-style score with a 0–5 turn score would silently produce garbage in the summary.

Two reasonable fixes:
- Validate at load time: reject duplicate metric names across scopes.
- Key by `(name, scope)` in `metric_scores` and label the summary row accordingly.

**3. A couple more stale references to "goal completion"**

- `evaluator.py:348-349` — `logger.error(f"Error computing goal completion for {convo_item.chat_id}: {e}")` — catches errors from `evaluate_conversation`, which now covers goal completion *plus* convo-scope custom metrics. Reword to "Error evaluating conversation".
- `evaluator.py:248` — the comment `# When auto: enough workers for all turns + goal_completion calls at once.` is now incomplete (also covers convo-scope custom metrics). The worker budget `num_workers = total_turns + len(conversations)` on line 251 also doesn't account for the `M` custom convo metrics running inside each `evaluate_conversation` call, though the inner pool is unbounded in auto mode so it works out — worth a comment.

arksim/config/core/agent.py
shehan360, line 1: rebase on the other pr and remove

shehan360 (changes requested):
Nice cleanup across the board — the refactor to a single `custom_metrics` list and the shared `_run_metrics_parallel` helper really tighten this up. A few more follow-ups:

**1. Restore `custom_qualitative_metrics` as a deprecated BC alias**

`EvaluationParams` is exported from `arksim/evaluator/__init__.py` and v0.3.4 shipped with a separate `custom_qualitative_metrics` field. Removing it means any user code like `EvaluationParams(custom_metrics=qs, custom_qualitative_metrics=qls)` silently drops the qualitative list on upgrade (Pydantic v2 defaults to `extra='ignore'`) — a worse failure mode than a loud error.

Simplest shim:

```python
from pydantic import model_validator
from warnings import warn

class EvaluationParams(BaseModel):
    custom_metrics: list[QuantitativeMetric | QualitativeMetric] = Field(
        default_factory=list
    )

    @model_validator(mode="before")
    @classmethod
    def _merge_legacy_qualitative_metrics(cls, data: Any) -> Any:
        if isinstance(data, dict) and "custom_qualitative_metrics" in data:
            warn(
                "`custom_qualitative_metrics` is deprecated; pass qualitative "
                "metrics in `custom_metrics` alongside quantitative ones.",
                DeprecationWarning,
                stacklevel=2,
            )
            legacy = data.pop("custom_qualitative_metrics") or []
            data["custom_metrics"] = list(data.get("custom_metrics") or []) + list(legacy)
        return data
```

Add a CHANGELOG entry under `### Deprecated`.

**2. `evaluate_conversation`: the goal-completion block is inconsistent with the rest of the function**

`evaluate.py:270-281`:

```python
if _should_run("goal_completion", metrics_to_run):
    result = GoalCompletionMetric(llm).score(
        ScoreInput(
            chat_history=convo_item.chat_history,
            user_goal=convo_item.user_goal,
        )
    )
    ...
```

Two smells:
- Rebuilds a fresh `ScoreInput` three lines after one was constructed at `evaluate.py:263-268`. Just reuse it.
- Runs sequentially before the parallel block, even though goal completion is another conversation-scope metric with its own LLM call. It could be included in the `_run_metrics_parallel` batch as just another convo-scope built-in, saving one round-trip of latency per conversation.

Minimal fix is reusing `score_input`; fuller fix is folding `GoalCompletionMetric` into the parallel batch and reading its result from the returned list.

**3. Scope routing in `evaluator.py:270-290` makes four passes over `custom_metrics`**

Four list comprehensions, each iterating the full list with an isinstance check and scope comparison. Collapse to a single pass:

```python
from collections import defaultdict

buckets: dict[tuple[str, str], list] = defaultdict(list)
for m in self.params.custom_metrics:
    kind = "quant" if isinstance(m, QuantitativeMetric) else "qual"
    buckets[(m.scope, kind)].append(m)

_turn_quant = buckets[("turn", "quant")]
_turn_qual = buckets[("turn", "qual")]
_convo_quant = buckets[("conversation", "quant")]
_convo_qual = buckets[("conversation", "qual")]
```

Metric lists are small so perf isn't the point — it's 16 lines of near-identical boilerplate collapsing to 6.

**4. `_run_quant` / `_run_qual` still reconstruct the result**

`evaluate.py:69-75`:

```python
def _run_quant(metric: QuantitativeMetric) -> QuantResult:
    r = metric.score(
        score_input.model_copy(update=getattr(metric, "additional_input", {}))
    )
    return QuantResult(
        name=metric.name, value=r.value, reason=r.reason, scope=metric.scope
    )
```

The metric already returns a `QuantResult`. The only thing the wrapper adds is stamping `scope=metric.scope`. Cleaner:

```python
def _run_quant(metric: QuantitativeMetric) -> QuantResult:
    r = metric.score(
        score_input.model_copy(update=getattr(metric, "additional_input", {}))
    )
    r.scope = metric.scope
    return r
```

Or stamp it in the base class once:

```python
# in QuantitativeMetric / QualitativeMetric
def run(self, score_input: ScoreInput) -> QuantResult:
    r = self.score(score_input)
    r.scope = self.scope
    return r
```

...then `_run_metrics_parallel` just calls `metric.run(...)` — the wrapping functions disappear entirely.

shehan360 (dismissed):
Thanks for addressing the prior round, the BC shim, the goal-completion refactor, the bucket dispatch, and the `run()` template method all look good. Two follow-ups on the latest commit:

**1. `QualitativeMetric.run()` drops `additional_input` forwarding (behavior regression)**

`base_metric.py`:

```python
# QuantitativeMetric.run
def run(self, score_input: ScoreInput) -> QuantResult:
    r = self.score(score_input.model_copy(update=self.additional_input))
    r.scope = self.scope
    return r

# QualitativeMetric.run
def run(self, score_input: ScoreInput) -> QualResult:
    r = self.evaluate(score_input)          # no model_copy / additional_input
    r.scope = self.scope
    return r
```

The previous `_run_qual` closure did `metric.evaluate(score_input.model_copy(update=getattr(metric, "additional_input", {})))`. A qual subclass that set `self.additional_input = {...}` in its own `__init__` (the documented side door, exercised by `test_custom_qual_receives_additional_input`) used to get that dict merged into `ScoreInput.model_extra`. The new `run()` skips that step, and the test asserting it was deleted in this commit, so the regression is silent.

**2. No test coverage for the new deprecation shim**

`_merge_legacy_qualitative_metrics` in `entities.py` is the only thing keeping users on `custom_qualitative_metrics=...` from silently losing data. There is no test that:

- asserts a `DeprecationWarning` fires on the legacy kwarg, or
- asserts the legacy list is correctly merged into `custom_metrics` (especially when both kwargs are passed together).

A short `pytest.warns(DeprecationWarning)` test in `tests/unit/test_evaluator_class.py` would lock the contract in so a future refactor cannot quietly drop it.

shehan360 (approved):

PR #166, opened 2026-05-01
fix(evaluator): align builtin metric prompt template with format args

KathyLiu20:
## Summary
- Rename prompt template placeholders to match the kwargs passed by builtin metrics (`full_conversation` → `chat_history`, `goal` → `user_goal`, `conversation` → `chat_history`), fixing `KeyError`s when these metrics run.
- Add `metadata: dict[str, Any] | None` field to `QualResult` so qualitative results carry the same metadata channel as `QuantResult`.
- Refine `agent_behavior_failure` prompt definitions for "lack of specific information" and "false information" categories.

## Test plan
- [x] `pytest tests/unit/`
- [x] Run a builtin-metric eval end-to-end and confirm no `KeyError` from prompt formatting
- [x] Confirm `QualResult.metadata` round-trips through serialization

### DIFF
--- a/arksim/evaluator/tool_call_metrics.py
+++ b/arksim/evaluator/tool_call_metrics.py
@@ -61,7 +61,7 @@ class ToolCallBehaviorFailureMetric(AgentBehaviorFailureMetric):
                     "role": "user",
                     "content": tool_call_behavior_failure_user_prompt.format(
                         user_goal=score_input.user_goal,
-                        chat_history=format_chat_history(score_input.current_turn),
+                        chat_history=format_chat_history(score_input.chat_history),
                         knowledge=score_input.knowledge,
                         tool_calls=json.dumps(
                             [
--- a/arksim/evaluator/tool_call_metrics.py
+++ b/arksim/evaluator/tool_call_metrics.py
@@ -61,7 +61,7 @@ class ToolCallBehaviorFailureMetric(AgentBehaviorFailureMetric):
                     "role": "user",
                     "content": tool_call_behavior_failure_user_prompt.format(
                         user_goal=score_input.user_goal,
-                        conversation=format_chat_history(score_input.current_turn),
+                        chat_history=format_chat_history(score_input.current_turn),
                         knowledge=score_input.knowledge,
                         tool_calls=json.dumps(
                             [
--- a/arksim/evaluator/base_metric.py
+++ b/arksim/evaluator/base_metric.py
@@ -42,6 +42,7 @@ class QualResult(BaseModel):
     name: str
     value: str
     reason: str | None = None
+    metadata: dict[str, Any] | None = None
     scope: MetricScope = "turn"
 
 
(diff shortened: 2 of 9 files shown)


arksim/evaluator/tool_call_metrics.py
```diff
@@ -61,7 +61,7 @@ def evaluate(self, score_input: ScoreInput) -> QualResult:
                     "role": "user",
                     "content": tool_call_behavior_failure_user_prompt.format(
                         user_goal=score_input.user_goal,
-                        conversation=format_chat_history(score_input.current_turn),
+                        chat_history=format_chat_history(score_input.current_turn),
```
shehan360: @itsarbit what should be passed to tool_call_behavior_failure, turn or the whole chat history?

shehan360 (changes requested):
Please add regression tests for the new contract

This PR changes two contracts that downstream code depends on:

- built-in prompt templates use config-renderable field names like `chat_history` and `user_goal`
- `QualResult` now preserves `metadata`, matching `QuantResult`

I do not see test files changed in the PR. It would be useful to add coverage that:

- each built-in prompt formats with a generic `ScoreInput` context containing `chat_history`, `current_turn`, `knowledge`, `user_goal`, and `profile`
- `QualResult(metadata=...)` round-trips through `model_dump` / `model_validate`
- a custom qualitative metric can return metadata and have it preserved in `evaluate_turn` or `evaluate_conversation`

arksim/evaluator/tool_call_metrics.py
itsarbit: I'm leaning to use full chat_history. The metric currently differs from `agent_behavior_failure`, which passes the full history.

Reasoning:
  1. Tool selection can depend on what the agent already gathered. For example, the user may say "yes" to a question from two turns earlier. Without that context, the judge may miss the intent.
  1. Argument correctness often relies on values from earlier turns, like order IDs or user details shared mid-conversation.
  1. Call necessity depends on whether the agent already called the same tool earlier. The current turn alone can't answer that.

Cost: prompts get longer per turn, with O(N^2) total input tokens for an N-turn conversation. That seems worth it for accuracy in pre-deploy evals. We can truncate later if this becomes load-bearing.

Implication for this PR: this changes behavior beyond the placeholder rename. Two options:

1. Change the value here too (score_input.current_turn -> score_input.chat_history), so the rename matches the actual input.
2. Keep current_turn for now and rename the placeholder to current_turn or last_turn instead of chat_history. This keeps the PR limited to the format-arg fix and tracks the behavior change separately.

I'd take option 1 because it matches agent_behavior_failure and gives the judge enough context. I'm open to option 2 if the original code used current_turn for a reason I'm missing. WDYT?

shehan360: Option 1 makes sense given the reasoning points!

shehan360 (approved):

